PR #37, opened 2013-11-24
Support for multiple MX records with the same priority

tzetze:
According to the RFC (http://tools.ietf.org/html/rfc974), it is normal to have multiple MX records at the same priority. To support that, i made this small change on the MX record manifest.


roderickm:
Please overlook commit 338659b, it had additional changes from my last contribution. I synced my fork and created a cleaner one (f744b4d) of the same changes.

PR #41, opened 2014-04-04
El compatible

Sereinity:
Make the module compatible with CentOs, RHEL, …


roderickm:
@leonidlm If you only need to include a single instead of managing the entire file, perhaps the [file_line source in stdlib](https://forge.puppetlabs.com/puppetlabs/stdlib#resources) would do what you need.

PR #56, opened 2014-08-11
added a dnssec_validation parameter, to avoid the creation of dnssec-val...

Sanic:
...idation auto in the bind server options file (.e.g named.conf.options)

In my setup, i needed to delete the dnssec_validation from the named.conf.options file. Maybe this is also relevant for other people


solarkennedy:
I'm going to close this as lost to the sands of time. If you still need this feature please rebase and reopen.

PR #57, opened 2014-08-12
add option to control ipv6 listeners. 

cypherfox:
Will support only none and any for now. Also fix small bug in function and param names in zone definition


solarkennedy:

PR #58, opened 2014-08-15
add parameter to set 'forwarder only' option

cypherfox:


roderickm:
Looks as if a test for 'forward only' was included in https://github.com/ajjahn/puppet-dns/commit/bfead975f2ff981982864cf666b74c4b97ef8dde

PR #67, opened 2014-12-09
Add no-dns-sec option

fr3dm4n:
This makes it possible to disable dns-sec

### DIFF
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -22,6 +22,7 @@ define dns::server::options(
   $check_names_slave = undef,
   $check_names_response = undef,
   $allow_query = [],
+  $no_dns_sec = false,
 ) {
   $valid_check_names = ['fail', 'warn', 'ignore']
 
--- a/spec/defines/dns__server__options_spec.rb
+++ b/spec/defines/dns__server__options_spec.rb
@@ -111,6 +111,16 @@ describe 'dns::server::options', :type => :define do
     it { should contain_file('/etc/bind/named.conf.options').with_content(/allow-query/)  }
 
   end
+  context 'disable dns-sec option' do
+    let :params do
+      { :no_dns_sec => true }
+    end
+
+    it { should_not contain_file('/etc/bind/named.conf.options').with_content(/dnssec-validation auto/)  }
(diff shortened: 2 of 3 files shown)


lifo101:
+1 this. I need to disable dnssec in my bind config and don't want to hack the module. Please merge this PR.

manifests/server/options.pp
```diff
@@ -22,6 +22,7 @@
   $check_names_slave = undef,
   $check_names_response = undef,
   $allow_query = [],
+  $no_dns_sec = false,
```
solarkennedy: What if you named this just "dnssec_validation" and made it default to true? I think that would be easier to read than a double negative.

manifests/server/options.pp
```diff
@@ -22,6 +22,7 @@
...
 ) {
   $valid_check_names = ['fail', 'warn', 'ignore']
```
solarkennedy: van you add a validate_bool to catch typos?

manifests/server/options.pp
solarkennedy: Can you also add a docstring for this parameter? The style so far is not great, can you start conforming to the styleguide with something like:
```
# == Paramaters
#
# [*dnssec_validation*]
# Bool, defaults to true. If false disabled dnssec and dnssec-validation.
#

File: spec/defines/dns__server__options_spec.rb [hunk 3]
```diff
@@ -111,6 +111,16 @@
     it { should contain_file('/etc/bind/named.conf.options').with_content(/allow-query/)  }
 
   end
+  context 'disable dns-sec option' do
```
solarkennedy: Thank you for writing this!

For style can you add a newline between the contexts and remove the neline on line 122?
(to give each context a more "self-contained" feeling)

darkfoxprime:
Has this issue been taken care of with the recent `dns::server::options::dnssec_validation` parameter?  it takes a value of `yes`, `no`, or `auto`, to set the `dnssec-validation` named option, or `absent` to remove the `dnssec-validation` option completely (for older versions of bind that don't implement dnssec)

solarkennedy:
Correct. dnssec_validation=>absent should fix this.

PR #68, opened 2014-12-10
Use the new build env on Travis

joshk:
dedicated cpu cores, more ram, better networking, faster boot times


solarkennedy:
@joshk can you confirm if this is still relevant? It looks like this PR just kinda moves things around and chops off the .0 on 2.1 ?

joshk:
This is still relevant, and removing the `.0` means you test against the latest `2.1` version, instead of just `2.1.0`

solarkennedy:
Ah, gotcha.

PR #71, opened 2015-01-17
MX preference fix, unique alias, add tests

roderickm:
Fixes ajjahn/puppet-dns#37
 - adds all fields to alias: host, zone, pref, and data
 - any three of the four can be the same for multiple MX records
   under management, which would previously cause a duplicate alias
 - add test to confirm concat fragment is named as expected

Fixes ajjahn/puppet-dns#39 by writing the promised tests

Other changes of note:
Default preference is now 10. A pref of 0 is allowed, but if a
DNS administrator wanted to insert a more-preferable MX record,
she would be unable to do so until the previous record TTL expired.

Default host is now '@'. This allows for cleaner manifests because
the most common MX case is to create an exchanger for the entire
$ORIGIN. BIND will do the substitution automatically.

Blank labels are permitted in zone files, but are not handled by
this module because it does not render concat fragments in an
explicit order. Since BIND will substitute the last valid label
for any blank label, the resulting substitution would be
unpredictable.

### DIFF
--- a/manifests/record/mx.pp
+++ b/manifests/record/mx.pp
@@ -1,15 +1,42 @@
 # == Define: dns::record::mx
 #
-# Wrapper for dns::record to set an XM record.
+# Wrapper for dns::record to set an MX record.
 #
 define dns::record::mx (
   $zone,
   $data,
-  $ttl = '',
-  $preference = '0',
-  $host = $name ) {
+  $ttl        = '',
+  $preference = 10,
+  $host       = '@' ) {
 
-  $alias = "${host},MX,${zone}"
+  $alias = "${host},${zone},MX,${preference},${data}"
+
+  validate_string($zone)
+  validate_string($data)
+  validate_string($host)
+
+  if !is_domain_name($zone) or $zone =~ /^[0-9\.]+$/ {
+    fail("Define[dns::record::mx]: MX zone ${zone} must be a valid domain name.")
+  }
+  # Highest label (top-level domain) must be alphabetic
+  if $zone =~ /\./ and $zone !~ /\.[A-Za-z]+$/ {
+    fail("Define[dns::record::mx]: MX zone ${zone} must be a valid domain name.")
+  }
+  # RR data must be a valid hostname, not entirely numeric values
+  if !is_domain_name($data) or $data =~ /^[0-9\.]+$/ {
+    fail("Define[dns::record::mx]: MX data ${data} must be a valid hostname.")
+  }
+  if !is_integer($preference) or $preference < 0 or $preference > 65536 {
+    fail("Define[dns::record::mx]: preference ${preference} must be an integer within 0-65536.")
+  }
+  if !is_domain_name($host) and $host != '@' {
+    # Blank labels are permitted in BIND zone files, but they are not handled
+    # by this puppet module because it does not render concat fragments in an
+    # explicit order. Since BIND will substitute the last valid label, the
+    # resulting blank substitution would be unpredictable.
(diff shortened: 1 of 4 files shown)


manifests/record/mx.pp
```diff
@@ -1,15 +1,36 @@
...
-  $preference = '0',
-  $host = $name ) {
+  $ttl        = '',
+  $preference = 10,
+  $host       = '@' ) {
 
-  $alias = "${host},MX,${zone}"
+  $alias = "${host},${zone},MX,${preference},${data}"
+
+  unless is_string($zone) and is_domain_name($zone) and !is_ip_address($zone) and $zone !~ /\.?\d+$/ {
```
solarkennedy: Wow these are fancy! Eventually maybe we need to build up our own validate_ttl functions or something :)

solarkennedy: Something about line 14 isn't compatible with puppet 2.7?
https://travis-ci.org/ajjahn/puppet-dns/jobs/47328674

The validation may be a bit overkill. How about just "is_domain_name()"? Shouldn't that cover it?

manifests/record/mx.pp
```diff
@@ -1,15 +1,36 @@
...
+    fail("Define[dns::record::mx]: MX zone ${zone} must be a valid domain name.")
+  }
+  unless is_string($data) and is_domain_name($data) and !is_ip_address($data) {
+    fail("Define[dns::record::mx]: MX data ${data} must be a valid hostname.")
+  }
+  unless $ttl =~ /^[0-9SsMmHhDdWw]+$/ or $ttl == '' {
```
roderickm: This TTL test is not specific to MX records. It should validate all BIND records, so would be better placed in the dns::record define.

manifests/record/mx.pp
roderickm: I agree that is_domain_name() should cover it, except that function from puppetlabs-stdlb currently returns true for IP addresses, and for numeric top-level domains, which are never domain names. This prompted me to add the !is_ip_address condition, the not-a-digit regex, and open [PUP-3856](https://tickets.puppetlabs.com/browse/MODULES-1670). Once is_domain_name is more strict, the last two conditions here could be dropped.

manifests/record/mx.pp
solarkennedy: Ah yes.

manifests/record/mx.pp
solarkennedy: Hmm. That's fine. I'm mostly worried about the failing test and would be "ok" to me if you dropped to the strictness to make it pass. Or just figure out why it is failing. (maybe it doesn't like the "!"?)

roderickm:
I'm stumped on why checks are failing for PUPPET_VERSION="~> 2.7.0". Rather than continuing archaeological whack-a-mole, I'll wait to see if someone can explain why '=~' and 'and' are causing syntax errors in record.pp:20 and record/mx.pp:14.

danzilio:
`unless` is not available in 2.7

roderickm:
Thanks, @danzilio. The Ruby Style Guide says I should [favor `unless` over `if` for negative conditions](https://github.com/bbatsov/ruby-style-guide#unless-for-negatives). Would this project prefer an `if not` conditional for 2.7 compatibility?

solarkennedy:
2.7 is still pretty widely deployed. I say yes on "if not".

roderickm:
Squashed and rebased for your review.

solarkennedy:
Super!

PR #72, opened 2015-01-26
allow port to be customized in dns::server::options

talisto:
I've added a `$listen_on_port` option to `dns::server::options` to allow the port to be customized.  Rspec tests included.

### DIFF
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -20,6 +20,7 @@
 define dns::server::options(
   $forwarders = [],
   $listen_on = [],
+  $listen_on_port = undef,
   $allow_recursion = [],
   $check_names_master = undef,
   $check_names_slave = undef,
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -8,7 +8,8 @@
 # $listen_on:
 #   Array of IP addresses on which to listen. Default: empty, meaning "any"
 # $listen_on_port:
-#   UDP/TCP port number to use for receiving and sending traffic.  Default: undefined, meaning 53
+#   UDP/TCP port number to use for receiving and sending traffic.
+#   Default: undefined, meaning 53
 # $group:
 #   Group of the file. Default: bind
 # $owner:
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -7,6 +7,8 @@
 #   Array of forwarders IP addresses. Default: empty
 # $listen_on:
 #   Array of IP addresses on which to listen. Default: empty, meaning "any"
+# $listen_on_port:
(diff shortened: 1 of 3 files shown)


templates/named.conf.options.erb
```diff
@@ -22,9 +22,9 @@ options {
 <% end -%>
 
 <% if @listen_on.size == 0 then -%>
```
solarkennedy: Hmm, my only problem here is this gate. You can't set the port without setting a listen_on?

manifests/server/options.pp
```diff
@@ -20,6 +20,7 @@
 define dns::server::options(
   $forwarders = [],
   $listen_on = [],
+  $listen_on_port = undef,
```
solarkennedy: Needs docstring at least.
We haven't been good at this, but can you add it per the format specified here?
https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc

templates/named.conf.options.erb
solarkennedy: I'm ok with that I guess, as long as it is documented.

talisto: Nope, the port can only be specified in the `listen-on` statement.  See here: http://docstore.mik.ua/orelly/networking_2ndEd/dns/ch10_15.htm

My other thought was to allow a hash for the `$listen_on` argument, like so:

```
$listen_on = {
'10.0.0.10' => 5300
}
```
..since in BIND 9 you can actually set the port on a per IP basis, but there'd still need to be a `$listen_on_port` option for when no `$listen_on` addresses are defined (i.e. `any`).

manifests/server/options.pp
talisto: ok, will do.  the docs for `dns::server::options` definitely need some other work too, though, as there are a bunch of options that aren't documented (`$allow_recursion`, `$check_names_master`, `$check_names_slave`, `$check_names_response`, `$allow_query`) and docs for some options that don't actually exist there (`$group`, `$owner`).

templates/named.conf.options.erb
talisto: oh, I just looked at the [official docs](https://kb.isc.org/article/AA-01031) and there actually is a `port` option, separate from the `listen-on` option.  I'll update the template to use that instead.

manifests/server/options.pp
talisto: I've created a separate pull request to fix up the other documentation for this class.  If you want to merge that one first, I'll rebase this and make the appropriate changes to match the docs in the other pull request.

solarkennedy:
Great!

PR #73, opened 2015-01-26
cleanup of inline rdocs in `dns::server::options` class

talisto:
Using style guide at https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc


solarkennedy:
Perfect. We should have been insisting on this from the beginning.
I guess I could have merged this first, then you probably wouldn't have to rebase, but can you fix the new conflict?

solarkennedy:
Thanks!

PR #77, opened 2015-02-10
Bind stats

gcmalloc:
adding support for configuring bind stat api

### DIFF
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -90,11 +90,11 @@ define dns::server::options (
   }
   validate_array($allow_query)
 
-  if $statistic_channel_port and !is_numeric($statistic_channel_port) {
+  if $statistic_channel_port != undef and !is_numeric($statistic_channel_port) {
     fail('The statistic_channel_port is not a number')
   }
 
-  if $statistic_channel_ip and (!is_string($statistic_channel_ip) or !is_ip_address($statistic_channel_ip)) {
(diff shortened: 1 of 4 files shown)


manifests/server/options.pp
```diff
@@ -70,6 +84,14 @@
   }
   validate_array($allow_query)
 
+  if !is_numeric($statistic_channel_port) {
+    fail('The statistic_channel_port is not a number')
+  }
+
+  if !validate_ip($statistic_channel_ip) {
```
solarkennedy: I think validate_ip auto fails, like validate_array or whatever, The extra if is extra

templates/named.conf.options.erb
```diff
@@ -62,6 +62,12 @@ check-names slave <%= @check_names_slave %>;
 check-names response <%= @check_names_response %>;
 <% end -%>
 
+<% if @statistic_channel_ip and @statistic_channel_ip -%>
```
solarkennedy: This needs a test.

Can you check out the spec/ folder and find similar tests?

I would like to see that the inet bla lport bla line is not in there by default, but is in there when you set the right ip/port. I wouldn't bother testing validation.

gcmalloc: ok, wanna have test for the other validation too ? I could add them in another pull request.

solarkennedy: Na, I don't think we need to test the validation. I'm mostly concerned about the template string values that end up on disk.

gcmalloc:
Sorry for the delay, should be good now.

solarkennedy:
Looks great! Thank you for the follow through!

Issue #78, opened 2015-02-13
Error finding a dependency.


robertdebock:
I see an error when using this module. My manifest is like this: (I have replaced sensitive information)
/etc/puppet/manifests/site.pp:
```
node 'hostname.domain.tld' inherits default {
 include dns::server
 include custom::zones
}
```

/etc/puppet/modules/custom/manifests/zones.pp:
```
class custom::zones {
  dns::zone { 'example.com':
   soa         => 'ns1.example.com',
   soa_email   => 'admin.example.com',
   nameservers => ['ns1.example.com']
  }
  dns::zone { '1.168.192.IN-ADDR.ARPA':
   soa         => 'ns1.example.com',
   soa_email   => 'admin.example.com',
   nameservers => ['ns1.example.com']
  }
  dns::record::a { 'ns1':
   zone => 'example.com',
   data => ['192.168.1.1'],
   ptr  => true;
  }
}
```

The error displayed is:
```
Error: Failed to apply catalog: Could not find dependent Exec[concat_/etc/named/zones/db.example.com.stage] for File[/var/lib/puppet/concat/_etc_named_zones_db.example.com.stage/fragments/9_db.example.com.ns1,A,example.com.record] at /etc/puppet/modules/concat/manifests/fragment.pp:123
```

My configuration:
```
puppet module list
/etc/puppet/modules
├── ajjahn-dns (v1.1.0)
├── puppetlabs-apache (v1.2.0)
├── puppetlabs-apt (v1.7.0)
├── puppetlabs-concat (v1.1.2)
├── puppetlabs-firewall (v1.4.0)
├── puppetlabs-inifile (v1.2.0)
├── puppetlabs-ntp (v3.3.0)
├── puppetlabs-postgresql (v4.1.0)
├── puppetlabs-puppetdb (v4.1.0)
├── puppetlabs-stdlib (v4.5.1)
├── saz-ssh (v2.4.0)
├── saz-sudo (v3.0.9)
├── stephenrjohnson-puppet (v1.3.1)
├── torrancew-account (v0.0.5)
└── custom (???)
/usr/share/puppet/modules (no modules installed)
```

My version of puppet:
```
puppet --version
3.7.4
```

PR #80, opened 2015-02-15
Added integration test with a full example host on redhat. Addresses #78

solarkennedy:
@robertdebock this adds a failing integration to cover exactly the bob exposed in #78.

However... I don't know what the deal is. It works if I set the facts to "debian".

It is like concat might be having trouble with the comma in the filename? I'm really not sure.


ITBlogger:
This can be closed

PR #84, opened 2015-02-18
fixed params.pp for rhel 7 and added fixes for concat issues

ITBlogger:


solarkennedy:
Obviously we were missing test coverage to catch the concat bug.

Can you take the tests from https://github.com/ajjahn/puppet-dns/pull/80/files
and add them into this PR to prove that this change actually fixes the issue?

Additionally travis is complaining about style guide violations.

ITBlogger:
Yeah, I'll add that stuff in...and fix the style issues...what's funny is that some of them are from the current build.

ITBlogger:
I also found some other issues that need to be addressed, so I'll get those added too. For instance, bind's config file needs includes for the local.conf and local.options.conf

solarkennedy:
Ok, try to limit the PR to one issue per PR, that makes it easy to ship and reason about changes.

Style guide violations... shouldn't be on the current build. Jenkins says the current build is passing. Are you on the same master as upstream?

ITBlogger:
Yes...I just know that I didn't make any changes to key.pp and I just
checked...it should be failing on the current build. Lines 64 and 65 have
one more space than is needed for the arrows.

--Alex

solarkennedy:
As far as I can tell, the current master (59db4a4cec1b3887ff52bfa46339ed627c7aa346) conforms to the style guide:
https://github.com/ajjahn/puppet-dns/blob/master/manifests/key.pp#L63-L65

Please pull master to make sure you are 100% up to date.

ITBlogger:
The current master does not conform to style guide...currently it's:
```
  concat::fragment { "${name}.key-secret":
    order   => 2,
    source  => "${cfg_dir}/bind.keys.d/${name}.secret",
  }
```
and should be:

```
  concat::fragment { "${name}.key-secret":
    order  => 2,
    source => "${cfg_dir}/bind.keys.d/${name}.secret",
  }
```

ITBlogger:
By the way, the problem with concat wasn't a bug in concat, but rather a problem with the puppet manifests. The manifests were written to dump files to a directory that didn't exist.

ITBlogger:
Yay! The build passed!

spec/hosts/example_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
+
+describe 'testhost.example.com' do
+
+  let(:facts) {{ :osfamily => 'RedHat', :concat_basedir  => '/dne', :define_fact => "" }}
+
+  context 'When given connected records that depend on each other' do
+    it { should compile }
```
solarkennedy: This test isn't testing anything yet because testhost.example.com is not yet defined.

Can you copy in the file from
https://github.com/ajjahn/puppet-dns/pull/80/files#diff-4f9b2b877cece59826df5b65a4980421R1

Adding that file and ensuring it can compile will be a definitive test that this all integrates correctly and the dependency bug is fixed, and will stay fixed.

ITBlogger:
Yeah, LOL, spec/fixtures was in .gitignore so it didn't get added to the previous commits

ITBlogger:
Yay, the build passed!

solarkennedy:
Great! Thank you very much. You have solved the concat issue and ensured that it won't happen again!

solarkennedy:
And lint fixes!

PR #85, opened 2015-02-20
Baseconfig

ITBlogger:
Added in control for named.conf and modified documentation on options to be specific to debian and rhel


solarkennedy:
This looks like a superset of the other PR. Can you rebase to reduce the diff?

PR #86, opened 2015-02-21
added control of named.conf and added documentation for options of both

ITBlogger:
debian and rhel


ITBlogger:
Just had to do a new pull request

templates/named.conf.erb
```diff
@@ -0,0 +1,26 @@
...
+// info UP in severity (no debug)
+// defaults to use 3 files in rotation
+// BIND 8.x logging MUST COME FIRST in this file
+// BIND 9.x parses the whole file before using the log
+// failure messages up to this point are in (syslog) 
+// typically /var/log/messages
+//
+  logging{
+  channel example_log{
+   file "<%= @log_file %>" versions 3 size 2m;
```
solarkennedy: This needs a test to ensure that it works and continues to work.

I'll help you write one:

```
context "When given a specific log location" do
  let(:params) {{ :log_dir => '/test_log' }}
  it { should contain_file('..named.conf').with_content(/file "\/test_log\/named\.log" version 3 size 2m/) }
end
```

manifests/server/config.pp
```diff
@@ -8,6 +8,9 @@
   $group    = $dns::server::params::group,
 ) inherits dns::server::params {
 
+  $log_dir  = '/var/log/named'
```
solarkennedy: How about putting this as a param? That way it can be a variable that can be set?

Is it not /var/log/bind on debian systems?

manifests/server/config.pp
```diff
@@ -28,12 +31,34 @@
     group  => $group,
     mode   => '0755',
   }
+  
+  file { '/var/cache/bind':
```
solarkennedy: Is this distro specific?

ITBlogger:
I'll work on this stuff soon

PR #87, opened 2015-03-23
Fixed a bug where key did not work on redhat due to incorrect pkg name

fhaynes:
Replaced the hardcoded package name with the package variable.

### DIFF
--- a/manifests/key.pp
+++ b/manifests/key.pp
@@ -15,7 +15,7 @@ define dns::key {
     command     => "/usr/sbin/dnssec-keygen -a HMAC-MD5 -r /dev/urandom -b 128 -n USER ${name}",
     cwd         => "${cfg_dir}/bind.keys.d",
     require     => [
-      Package['dnssec-tools','bind9'],
+      Package['dnssec-tools',$package],
       File["${cfg_dir}/bind.keys.d"],
     ],
     refreshonly => true,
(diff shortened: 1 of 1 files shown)


manifests/key.pp
```diff
@@ -15,7 +15,7 @@
     command     => "/usr/sbin/dnssec-keygen -a HMAC-MD5 -r /dev/urandom -b 128 -n USER ${name}",
     cwd         => "${cfg_dir}/bind.keys.d",
     require     => [
-      Package['dnssec-tools','bind9'],
+      Package['dnssec-tools',$package],
```
solarkennedy: I don't understand, where does this variable come from? I don't see it in the namespace?

fhaynes: It is from dns::server::params I believe.

solarkennedy: So.. shouldn't it be dns::server::package or something?

Alternatively I would argue that this should just require 'dnssec-tools' and let the other classes require their own packages?

fhaynes: My apologies, when I was preparing a clean branch for the pull request I neglected to include the $package = $dns::server::params::package line in the file, so you're right, $package would have been undef. I pushed a new commit that fully qualifies it, opting to just fully qualify it in the Package require.

Regarding just requiring dnssec-tools, that is probably a better choice. I was just doing the minimum necessary to fix the bug and get it working on RedHat while attempting to preserve the function of the original code. If you'd rather go that route, I'll push a commit that removes the requirement for the bind package entirely.

solarkennedy: Yea, the dnssec-tools is the actual requirement, as attached to the exec. It doesn't actually need to depend on bind.

PR #88, opened 2015-03-23
Fixed a bug where the key was being written with }: and not };

fhaynes:
This was causing bind to not start


solarkennedy:
Thanks! I'm not sure how this got past...

PR #90, opened 2015-03-23
Added in feature allowing for global allow-transfer

fhaynes:
Would you guys find this useful? It looked like existing allow-transfer feature was for the zone level only. We have some need for global allow-transfer, so I added this in. Based the unit tests on the forwarders tests.

### DIFF
--- a/spec/defines/dns__server__options_spec.rb
+++ b/spec/defines/dns__server__options_spec.rb
@@ -23,12 +23,36 @@ describe 'dns::server::options', :type => :define do
 
   end
 
+  context 'passing valid array to transfers' do
+    let :params do
+      { :transfers => ['192.168.0.3', '192.168.0.4'] }
+    end
+
+    it { should contain_file('/etc/bind/named.conf.options') }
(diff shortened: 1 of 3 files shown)


spec/defines/dns__server__options_spec.rb
```diff
@@ -23,6 +23,21 @@
...
 
+  context 'passing valid array to transfers' do
+    let :params do
+      { :transfers => ['192.168.0.3', '192.168.0.4'] }
+    end
+
+    it { should contain_file('/etc/bind/named.conf.options') }
+    it { should contain_file('/etc/bind/named.conf.options').with_content(/192\.168\.0\.3;$/) }
+    it { should contain_file('/etc/bind/named.conf.options').with_content(/192\.168\.0\.4;$/) }
+    it { should contain_file('/etc/bind/named.conf.options').with_ensure("present") }
```
solarkennedy: This is a tad over specified I think. The important part to me is the content

solarkennedy:
This is excellent!

PR #91, opened 2015-03-23
Added in a feature to allow zones files to have supplementary files

fhaynes:
In our environment some of our records are generated by a script with data pulled from a database, and it writes these out to files BIND then reads. There seemed to be no easy way to support that workflow with this module, so I added in the idea of a supplementary file.

The user can specify that a zone will have a supplementary file, and if so, it just adds an include directive (and deploys a blank file to avoid BIND errors) to include the file, and then you can overwrite that file with whatever records desired.

I am not sure if you think this is appropriate for the general use module or useful for your users, but I figured I would give you a pull request for it in case you wanted to include it.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -16,7 +16,8 @@ define dns::zone (
   $forward_policy = 'first',
   $slave_masters = undef,
   $zone_notify = false,
-  $ensure = present
+  $ensure = present,
+  $supplementary = false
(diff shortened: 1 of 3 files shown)


solarkennedy:
I understand why you have this, but with the other PR that allows "updates" which can just set replace=>false, doesn't that achieve the same effect?

PR #92, opened 2015-03-23
Added in support for allow-update for zones

fhaynes:
I needed support for allowing a zone to be updated dynamically, so I put this in. I made sure that the module will not overwrite a file while it is allowing updates from an IP, as that would overwrite the dynamic leases. Also preventing it from bumping the zone serial while it is accepting updates.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -13,6 +13,7 @@ define dns::zone (
   $zone_type = 'master',
   $allow_transfer = [],
   $allow_forwarder = [],
+  $allow_update = undef,
   $forward_policy = 'first',
   $slave_masters = undef,
   $zone_notify = false,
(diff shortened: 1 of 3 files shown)


manifests/zone.pp
```diff
@@ -13,6 +13,7 @@
   $zone_type = 'master',
   $allow_transfer = [],
   $allow_forwarder = [],
+  $allow_update = undef,
```
solarkennedy: Can you do a plain bool here?

manifests/zone.pp
```diff
@@ -46,13 +47,24 @@
   } else {
     # Zone Database
 
+    # If it is a dynamic zone, we don't want to overwrite the existing zone file as it might
+    # destroy records that have been added
+    if $allow_update {
```
solarkennedy: Can you validate_bool() on this before you use it?

manifests/zone.pp
solarkennedy: It looks like we haven't been adding docstrings to this type.

Can you start with
```
# == Parameters
#
# [*allow_update*]
# ...
#
define dns::zone...
```

fhaynes: If I do a bool there, I will need a second parameter to hold the list of IPs allowed to update. validate_bool() will fail if an array is passed to it, I think. If the second variable is acceptable, I'll push a commit with that change.

solarkennedy: Ah, gotcha. Eh this is fine.

manifests/zone.pp
solarkennedy: Then what type are you expecting after the if block? Can you validate then?

solarkennedy: Array I suppose, right.

solarkennedy:

PR #99, opened 2015-04-10
zone files should only be created or modified for master zones

jearls:
The `zone.pp` class should not be allowed to create or modify the zone file for a slave zone at all.  It should only manage master zone files.

I just added in a check to `zone.pp` so it only creates the `concat` type for master files, for all others it sets the `concat` to `ensure => absent`.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -43,7 +43,7 @@ define dns::zone (
     file { $zone_file:
       ensure => absent,
     }
-  } else {
+  } elsif $zone_type == 'master' {
     # Zone Database
 
     # Create "fake" zone file without zone-serial
@@ -73,6 +73,12 @@ define dns::zone (
       group       => $dns::server::params::group,
       require     => Class['dns::server::install'],
       notify      => Class['dns::server::service'],
+    }
+  } else {
+    # For any zone file that is not a master zone, we should make sure
+    # we don't have a staging file
+    concat { $zone_file_stage:
+      ensure => absent
     }
   }
 
--- a/spec/defines/dns__zone_spec.rb
+++ b/spec/defines/dns__zone_spec.rb
@@ -104,7 +104,7 @@ describe 'dns::zone' do
           end
           it 'should not have allow_tranfer entry' do
               should_not contain_concat__fragment('named.conf.local.test.com.include').
-                         with_content(/allow_transfer/)
+                         with_content(/allow-transfer/)
           end
           it 'should not have file entry' do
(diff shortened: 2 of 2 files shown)


jearls:
Sorry, clicked create too quickly.

The `zone.pp` class should not be allowed to create or modify the zone file for a slave zone at all.  It should only manage master zone files.

I just added in a check to `zone.pp` so it only creates the `concat` type for master files, for all others it sets the `concat` to `ensure => absent`.

solarkennedy:
Hmm, can you add some tests to cover these cases? As this manifest gets more complicated, we'll need test coverage to prevent regressions.

jearls:
I can try.  I have yet to actually write a test case for puppet :)

solarkennedy:
I'll help you!

See the CONTRIBUTING.md for initial instructions.

Your file hint is spec/defines/dns__zone_spec.rb

It might look like something like this:

```ruby
context "On a master zone" do
  let(:params) {{ :zone_type => 'master' }}
  it { should contain_concat__fragment ....
end

context "On a slave zone" do
  let(:params) {{ :zone_type => 'slave' }}
  it { should contain_concat__fragment(..).with_ensure('absent')
end
```

Tell me again though why on a slave zone we need this change? Certainly on a slave zone it shouldn't have an actual zone file, but it should have the configuration file fragment to declare the zone itself, right?

jearls:
actually, I think I've got what I need working (including some fixes to the existing zone spec file).

The reason the change is needed is that the bind server manages the zone file for a slave zone by itself, and that zone file should never be modified other than by the bind server.

the change I made only means that the `$zone_file_stage` concat is set to `ensure => absent` for any non-master zone; it doesn't affect creation of the zone config.

jearls:
... well, i did have things working, until i merged the latest changes back in so i could stop getting the some of the test file warnings and now all the requirements have changed and i need to rebuild ruby to a newer version on my system.  oh well.

jearls:
Woot.  Figured out how to revert the merge, ran the tests (and ignored the warning about the invalid regexp), and everything passed.  Checked in the latest changes.  Do I need to do anything special like try to get all the changes in one commit?  (This is also my first time contributing to an open source project like this)

solarkennedy:
As far as I know, this is correct.
Thank you for going all the way and writing tests to ensure that this continues to work. (I can easily see the case statement being refactored, etc.)

solarkennedy:
Thanks!

solarkennedy:
Strange, this seemed to pass tests in the PR but now is failing on master...

PR #100, opened 2015-04-11
Use resource names instead of hosts for the aliases of dns record types.  With spec test file.

jearls:
I noticed that the current resource definitions in the `dns::record::`* classes all base their resource names on the zonefile host entry.  This means that you cannot create multiple entries of the same type for the same name.

Admittedly,most of the entries allow for an array for the data parameter, but there's one case where this doesn't work:  automatic ptr entry creation.

Suppose I want to create these DNS zones fragments:

    foo     IN A    192.168.100.101
    foo     IN A    192.168.100.102

...and...

    101    IN PTR    foo.example.com.
    102    IN PTR    foo.example.com.

Under the current code, I would not be able to use the automatic PTR record generation for this, because I would have to use a two-item array for the data of the A record, but the automatic PTR generation only looks at the first IP address.

This PR changes the alias definitions for the record type to use `${name}` instead of `${host}`.  This would allow me to give the two A records different names but use the same host entry in both, and I would get my automatic PTR generation:

    dns::record::a { "foo,101":    # generates alias foo,101,A,example.com
        host => "foo" ,
        zone => "example.com" ,
        data => [ "192.168.100.101" ] ,
        ptr => true
    }
    
    dns::record::a { "foo,102":    # generates alias foo,102,A,example.com
        host => "foo" ,
        zone => "example.com" ,
        data => [ "192.168.100.102" ] ,
        ptr => true
    }

This also allows, in general, for the (to me) more obvious style of having two resource definitions if I want to create two records of the same name, whether it's for A records, NS records, TXT records, etc.

MX records already allow this, but I made the change to them also to be consistent.  The only record type I did not change was the SRV record type, since I don't understand the semantics of the SRV records.

I also created a spec file to test all the record types that use the new alias construction, to ensure they create the proper fragment name and proper record entry both when using the alias for the host and when specifying a different host, and I adjusted the `dns__record__mx_spec.rb` tests to match the new expected fragment names.

### DIFF
--- a/manifests/record/a.pp
+++ b/manifests/record/a.pp
@@ -10,7 +10,7 @@ define dns::record::a (
   $ptr = false,
   $host = $name ) {
 
-  $alias = "${host},A,${zone}"
+  $alias = "${name},A,${zone}"
 
   dns::record { $alias:
     zone => $zone,
--- a/manifests/record/aaaa.pp
+++ b/manifests/record/aaaa.pp
@@ -8,7 +8,7 @@ define dns::record::aaaa (
   $ttl = '',
   $host = $name ) {
 
-  $alias = "${host},AAAA,${zone}"
+  $alias = "${name},AAAA,${zone}"
 
   dns::record { $alias:
     zone   => $zone,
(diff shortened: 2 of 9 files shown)


solarkennedy:
Thanks!

Issue #101, opened 2015-04-15
named.conf not including options


chriscowley:
Bind is loading the options in `/etc/named.conf` with no problems. However, it is only listening on localhost:53 (for example) as defined in that file.

The module creates `/etc/named/named.conf.local` and `/etc/named/named.conf.options`, but from what I can see this is not included in /etc/named.conf (which does not appear to be touched by the module).

Net result: nothing happens other than Binding as a caching nameserver accessible to localhost.

solarkennedy:
I think we (by default) leave /etc/named.conf alone, I think most distros include more from there?
What distro are you on?

chriscowley:
Centos 6 - it does not include /etc/named/ by default. Is there a way I can force your module to purge that file, or do I need to modify it myself (started) and submit a PR?

solarkennedy:
```
file { '/etc/named.conf': ensure => absent }
```
But no, currently the module does not manage that file.

If you think this module should manage that file, then PR please. 

But per params.pp:
https://github.com/ajjahn/puppet-dns/blob/master/manifests/server/params.pp#L17

It looks like on RedHat distros we *do* manage it?

chriscowley:
I saw the entry in params.pp, but it was most certainly leaving the main conf file alone. Its not just me having a special moment (which did happen only yesterday), one of my colleagues has confirmed it.

I've got everything working in our environment, I'm just cleaning it all up make it acceptable for a PR.

jearls:
The dns::server::config module references the config file:

      file { $cfg_file:
        ensure  => present,
        owner   => $owner,
        group   => $group,
        mode    => '0644',
        require => [
          File[$cfg_dir],
          Class['dns::server::install']
        ],
        notify  => Class['dns::server::service'],
      }

Unfortunately, because of this, it also means we can't fix the real `named.conf` file elsewhere in puppet because it's already defined here.  The RedHat version of `named.conf` is automatically set up to be a local-caching-only server, which obviously won't work for a real dns server.  Is there any progress on the named.conf PR?

ppouliot:
Has this been fixed yet.  looks like named.conf.options doesn't get created when used on RH based systems (RH, Centos, Fedora, ScientificLinux).

solarkennedy:
I believe this is fixed thanks to #102 
https://github.com/ajjahn/puppet-dns/blob/6f649f66cc0b352332e418f08b11b710b9a59144/templates/named.conf.erb#L4

PR #102, opened 2015-04-16
Control named.conf

chriscowley:
Added support for controlling the content of `/etc/named` on RHEL as discussed in [issue 101](https://github.com/ajjahn/puppet-dns/issues/101)

I have tested this on CentOS 6 and I hope I have not created any regressions.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -29,16 +29,35 @@ class dns::server::config (
     mode   => '0755',
   }
 
-  file { $cfg_file:
-    ensure  => present,
-    owner   => $owner,
-    group   => $group,
-    mode    => '0644',
-    require => [
-      File[$cfg_dir],
-      Class['dns::server::install']
-    ],
-    notify  => Class['dns::server::service'],
+  case $::osfamily {
+    'RedHat': {
+      file { $cfg_file:
+        ensure  => present,
+        owner   => $owner,
+        group   => $group,
+        mode    => '0644',
+        content => template("${module_name}/named.conf.erb"),
+        require => [
+          File[$cfg_dir],
+          Class['dns::server::install']
+        ],
+        notify  => Class['dns::server::service'],
+      }
+    }
+    'Debian': {
+      file { $cfg_file:
+        ensure  => present,
+        owner   => $owner,
+        group   => $group,
+        mode    => '0644',
+        require => [
+          File[$cfg_dir],
+          Class['dns::server::install']
+        ],
+        notify  => Class['dns::server::service'],
+      }
+    }
+    default: { fail('Unrecognized operating system') }
   }
 
   concat { "${cfg_dir}/named.conf.local":
--- a/templates/named.conf.options.erb
+++ b/templates/named.conf.options.erb
@@ -1,6 +1,8 @@
 options {
 <% if @osfamily == 'Debian' -%>
   directory "/var/cache/bind";
+<% else -%>
+    directory "<%= @data_dir %>";
 <% end -%>
 	// If there is a firewall between you and nameservers you want
 	// to talk to, you may need to fix the firewall to allow multiple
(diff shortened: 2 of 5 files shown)


templates/named.conf.options.erb
```diff
@@ -1,6 +1,9 @@
 options {
-	directory "/var/cache/bind";
-
+<% if @osfamily == 'Debian' -%>
+  directory "/var/cache/bind";
+<% else -%>
+    directory "<%= @data_dir %>";
```
solarkennedy: Do we not have a data_dir on non-Debian oses? Can you make it not hard-coded in either case?

templates/named.conf.options.erb
```diff
@@ -73,7 +77,12 @@ check-names response <%= @check_names_response %>;
 	// If BIND logs error messages about the root key being expired,
 	// you will need to update your keys.  See https://www.isc.org/bind-keys
 	//========================================================================
-	dnssec-validation auto;
+	dnssec-validation yes;
```
solarkennedy: Can you save these changes for a different review? There is another PR for a similar thing.

manifests/server/config.pp
```diff
@@ -29,16 +29,35 @@
...
-    ],
-    notify  => Class['dns::server::service'],
+  case $::osfamily {
+    'RedHat': {
+      file { $cfg_file:
+        ensure  => present,
+        owner   => $owner,
+        group   => $group,
+        mode    => '0644',
+        content => template("${module_name}/named.conf.erb"),
```
solarkennedy: Is there any particular reason why we don't template out this file in both redhat and debian cases (assuming that all the OS specific differences are in the params.pp file?

That would greatly simplify the changes to this file, and hide any OS-specific paths in params.pp

jearls: The [debian `named.conf` file](https://wiki.debian.org/Bind9#named.conf_File) does not need to be replaced - it already works.  Mr. Crowley's redhat named.conf is basically implementing the same configuration as the debian version.

jearls: Mr. Crowley:  Your `dns__server__config_spec.rb` changes introduced a couple of typos:

On line 38, you need to add an additional `}` at the end of the line to close out the `it { …` block.
On line 39, `shoud` needs to be `should`, and I believe you need to enclose the should expression in `{…}` (e.g. `it { should contain_file… }`)

jearls: One other error on line `dns__server__config_spec.rb`:39 - the included file needs to be in `"`-quotes:  `….with_content("/^include \"/etc/named/named.conf.options\";$/")`

jearls: complete diff of changes to make your fork pass spec:

    diff --git a/spec/classes/dns__server__config_spec.rb b/spec/classes/dns__server__config_spec.rb
    index fc386e8..cc6ce82 100644
    --- a/spec/classes/dns__server__config_spec.rb
    +++ b/spec/classes/dns__server__config_spec.rb
    @@ -35,8 +35,8 @@ describe 'dns::server::config', :type => :class do
           'owner'  => 'named',
           'group'  => 'named',
           'mode'   => '0644',
    -    })
    -    it shoud contain_file('/etc/named.conf').with_content("/^include /etc/named/named.conf.options;$/")
    +    })}
    +    it { should contain_file('/etc/named.conf').with_content(/^include "\/etc\/named\/named.conf.options";$/) }
       end
     end

    diff --git a/templates/named.conf.erb b/templates/named.conf.erb
    index 3221629..10ffbb9 100644
    --- a/templates/named.conf.erb
    +++ b/templates/named.conf.erb
    @@ -1,7 +1,7 @@
     ; File managed by puppet.
     ;

    -include /etc/named/named.conf.options;
    +include "/etc/named/named.conf.options";
     logging {
             channel default_debug {
                     file "data/named.run";

I added the `"…"` around the `named.conf.options` include to match correct `named.conf` syntax.

jearls:
I submitted a new PR (https://github.com/ajjahn/puppet-dns/pull/112) that fixes the spec file and addresses @solarkennedy's concerns.

solarkennedy:
Closing in favor of #112

PR #103, opened 2015-04-19
Added file_concat as a dependent module

solarkennedy:
The build started failing after I merged in #99, but it was passing when I clicked the green button.

This is because the concat module has changed since then, and our fixtures are set to always build against master.

This PR adds file_concat as a dependent module, as apparently that is a thing now:
https://github.com/puppetlabs/puppetlabs-concat#what-concat-affects

At first I thought about just pinning our fixtures to released versions, which seems like a good short term fix. However it will blind us to real breaking changes, and also means we have to manually bump it all the time to test against new versions?

I'm up for alternatives. Obviously I want the built to pass. I'm surprised that concat has an external dependency now, especially a non-puppetlabs one? But if they depend on it, then we must depend on it too (eventually).

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,5 +2,6 @@ fixtures:
   repositories:
     stdlib: "git://github.com/puppetlabs/puppetlabs-stdlib.git"
     concat: "git://github.com/puppetlabs/puppetlabs-concat.git"
+    file_concat: "git://github.com/electrical/puppet-lib-file_concat.git"
   symlinks:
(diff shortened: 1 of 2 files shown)


jearls:
The author of the `file_concat` module has an [issue request](https://tickets.puppetlabs.com/browse/MODULES-1946) up with PuppetLabs to get his module marked as a "Puppet Approved" module.  I think that would help alleviate some of the concerns, at least.

PR #106, opened 2015-05-04
Update dns::server::params class to not install dnssec-tools for RedHat 6

jearls:
`dnssec-keygen` is included with the normal `bind` package in RedHat 6; there is no separate `dnssec-tools` package.


jearls:
Cancelling.  I should have made sure this syntax worked before submitting.  I will re-do and submit again later.

PR #107, opened 2015-05-05
Fix package list for newer versions of RedHat.

jearls:
Switch RedHat package conditional logic around so that the default is to only get bind, and RHEL versions 1-5 get bind and dnssec-tools.

### DIFF
--- a/manifests/server/params.pp
+++ b/manifests/server/params.pp
@@ -20,13 +20,7 @@ class dns::server::params {
       $owner              = 'named'
       $package            = 'bind'
       $service            = 'named'
-      case $::operatingsystemmajrelease {
-        /^[0-5]$/: {
-          $necessary_packages = [ 'bind', 'dnssec-tools' ]
-        }
-        default: {
-          $necessary_packages = [ 'bind', ]
-        }
+      $necessary_packages = [ 'bind' ]
       }
     }
     default: {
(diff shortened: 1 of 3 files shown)


jearls:
This PR has been tested on my local system.  It's needed because RHEL 6 does not have the `dnssec-tools` package, and so is failing in the `dns::server::install` class.

solarkennedy:
This is related to #84.

I doubt EL5 has dnssec-tools either.

I vote you just take out the dnssec-tools dep entirely. I think you need epel for it, and it isn't a hard requirement to run this module, so I think params should reflect that.

jearls:
Yes, it looks like it's part of epel, and you're right, it looks like dnssec-keygen is part of bind there too.  I'll rework the PR.

jearls:
Okay, removed dnssec-tools from the package list.  I also added a RedHat family validation to the spec/classes/dns__server__install_spec.rb test.

jearls:
whoops, failed spec.  let me try again.

jearls:
dangit.  it passed spec, i committed, then it failed spec.  let me check again.  sorry for the mess.

jearls:
i'm confused.  when i run "bundle exec rake spec", it's reverting some of my changes back to the master branch:

    $ bundle exec rake spec
    HEAD is now at 17dfb7e Merge pull request #99 from jearls/only-update-master-zone-files

What am I doing wrong?

jearls:
I'm going to withdraw this PR, update my forked repo to match the latest, the redo it.

solarkennedy:
I'm surprised that bundle exec rake spec is reverting changes.
It shouldn't?
https://github.com/ajjahn/puppet-dns/blob/master/.fixtures.yml#L7

PR #108, opened 2015-05-05
Remove dnssec-tools from RedHat package list.

jearls:
Also add RedHat validation to `dns__server__install_spec.rb`

### DIFF
--- a/spec/classes/dns__server__install_spec.rb
+++ b/spec/classes/dns__server__install_spec.rb
@@ -16,9 +16,7 @@ describe 'dns::server::install', :type => :class do
   context "on a RedHat OS" do
     let(:facts) {{ :osfamily => 'RedHat' }}
     it { should contain_class('dns::server::params') }
-    ['bind'].each do |package|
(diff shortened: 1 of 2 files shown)


spec/classes/dns__server__install_spec.rb
```diff
@@ -13,5 +13,13 @@
     end
   end
 
+  context "on a RedHat OS" do
+    let(:facts) {{ :osfamily => 'RedHat' }}
+    it { should contain_class('dns::server::params') }
+    ['bind'].each do |package|
```
solarkennedy: This is overkill. I know it looks like it does above, just kiss

```
it { should contain_package('bind') }
```

jearls:
Okay, I simplified the test.

solarkennedy:
Nice. I think this is fine.

Someday someone smart will add a whole dnssec thing, auto-gen keys, sign zones, etc. When that day comes we will have a toggle, and when on, will ensure that package is installed.

PR #110, opened 2015-05-05
Add notify to server options & also_notify to server and zone options.

jearls:
This pull request adds more complete notification options to both zones
and the server options.  In both cases, `zone_notify` sets the BIND
"notify" option (which can be one of *on*, *off*, *master-only*, or
*explicit*), and `also_notify` is a list of ip addresses of name servers
that will be sent notifications in addition to the zone's nameservers.

Both zone_notify and also_notify are validated in dns::server::options
and dns::zone, and new tests are added to dns__server__options_spec.rb
and dns__zone_spec.rb to test the individual values for each parameter.

### DIFF
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -49,6 +49,19 @@
 #            both statistic_channel_port and statistic_channel_ip must be defined
 #            for the statistic api to be enabled
 #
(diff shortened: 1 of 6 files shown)


solarkennedy:
Looks great!

PR #111, opened 2015-05-07
Added in support for /etc/default/bind9

Cicco0:
I needed support for accessing the startup hocks, so I put this in. Now you are able to define to resolve the configuration or to start bind in ipv4 / ivp6 mode.
I made sure that the module will not overwrite any other file.

### DIFF
--- a/manifests/server/params.pp
+++ b/manifests/server/params.pp
@@ -20,14 +20,7 @@ class dns::server::params {
       $owner              = 'named'
       $package            = 'bind'
       $service            = 'named'
-      case $::operatingsystemmajrelease {
-        '7': {
-          $necessary_packages = [ 'bind', ]
-        }
-        default: {
-          $necessary_packages = [ 'bind', 'dnssec-tools' ]
-        }
-      }
+      $necessary_packages = [ 'bind', ]
     }
     default: {
       fail("dns::server is incompatible with this osfamily: ${::osfamily}")
--- a/spec/classes/dns__server__install_spec.rb
+++ b/spec/classes/dns__server__install_spec.rb
@@ -13,5 +13,11 @@ describe 'dns::server::install', :type => :class do
     end
   end
 
+  context "on a RedHat OS" do
+    let(:facts) {{ :osfamily => 'RedHat' }}
+    it { should contain_class('dns::server::params') }
+    it { should contain_package('bind') }
+  end
+
 end
 


solarkennedy:
Is this something we need at all? If puppet is managing bind, can we just replace this file wholesale?
I don't think we need the complexity of a new defined type. (also, it is debian specific right?)

Cicco0:
You are right it is debian specific. I just saw that Redhat also offers such a configuration. I'll adjust it so that other system also supported.

> I don't think we need the complexity of a new defined type. (also, it is debian specific right?)
Damn, that does not work with other system, you're right. I push it again, once the adjustment is made.

solarkennedy:
Well debian has /etc/default/bind and redhat has /etc/sysconfig/named

I think it makes sense to do it for both distros that we support.

But it doesn't have to be a defined type, it can just be in the config class.

Cicco0:
Correct, I'm already on rewrite. Should be ready on Monday :)

Cicco0:
Rewriting was performed. It is now managed as a class and support multiple systems.
The separation is still present, as this is for the start of the service.

Cicco0:
Rewriting was performed. It is now managed as a class and support multiple systems.
The separation is still present, as this is for the start of the service.

manifests/server/params.pp
```diff
@@ -20,10 +23,20 @@
       $owner              = 'named'
       $package            = 'bind'
       $service            = 'named'
-      $necessary_packages = [ 'bind', ]
+      case $::operatingsystemmajrelease {
+        '7': {
+          $necessary_packages = [ 'bind', ]
+        }
+        default: {
+          $necessary_packages = [ 'bind', 'dnssec-tools' ]
```
solarkennedy: This un-does a PR that removed dnssec tools

manifests/server/params.pp
```diff
@@ -20,10 +23,20 @@
...
+        }
+      }
+      $prm_file           = "/etc/sysconfig/named"
+      $prm_template       = "startup.redhat.pp"
```
solarkennedy: Usually template files end in .erb?

manifests/server/params.pp
```diff
@@ -11,6 +12,8 @@
       $package            = 'bind9'
       $service            = 'bind9'
       $necessary_packages = [ 'bind9', 'dnssec-tools' ]
+      $prm_file           = "/etc/default/bind9"
+      $prm_template       = "startup.debian.pp"
```
solarkennedy: erb

manifests/server/params.pp
```diff
@@ -11,6 +12,8 @@
       $package            = 'bind9'
       $service            = 'bind9'
       $necessary_packages = [ 'bind9', 'dnssec-tools' ]
+      $prm_file           = "/etc/default/bind9"
```
solarkennedy: I've never heard the word "prm" before, what does it stand for?

solarkennedy:
Can you re-merge/rebase against master so the PR will be mergeable?

Cicco0:
The changes have been corrected. Thanks for the note.
PRM was replaced with startup, so it is recognizable.

solarkennedy:
Great! I think this is pretty clear now.

There are still 2 things that need to happen before this can be merged:
1. Needs to merge master or rebased
2. Needs at least the most basic tests to make sure it is syntactically correct and does what you think it does.

Here, I'll help you write one! (just typed in github, may be full of typos)

```ruby
# spec/classes/server/startup_spec.rb
require 'spec_helper'
describe 'dns::server::startup' do

  context 'by default on debian' do
    let(:facts) {{ :osfamily => 'Debian' }}
    it { should contain_file('/etc/default/bind9').with_content(/RESOLVCONF=no/) }
  end
  
  context 'by default on redhat' do
    let(:facts) {{ :osfamily => 'RedHat' }}
    it { should contain_file('/etc/sysconfig/named').with_content(/OPTIONS=something/) }
  end

end
```

I'm not as interested in double checking the redhat/debian thing, I'm more interested to make sure the files are correct, there are no typos, the erb synax is good, etc. These tests will ensure that always is the case, as it will exercise both templates.

Cicco0:
I'll start over with a new pull request. My repo is a little bit messy. Sorry, I'm a first timer.

PR #112, opened 2015-05-07
issue 101: take control of named.conf.

jearls:
This PR incorporates and fixes the changes proposed by chriscrowley
in PR https://github.com/ajjahn/puppet-dns/pull/102:

-   control content of named.conf in addition to simple presence and ownership.
    -   add to server params the additional named.conf settings that differ between Debian and RedHat
    -   use a single template for both OSes
-   add tests for RedHat and update tests for Debian to ensure that named.conf is getting created with proper ownership and content.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -34,6 +34,7 @@ class dns::server::config (
     owner   => $owner,
     group   => $group,
     mode    => '0644',
+    content => template("${module_name}/named.conf.erb"),
     require => [
       File[$cfg_dir],
       Class['dns::server::install']
--- a/templates/named.conf.erb
+++ b/templates/named.conf.erb
@@ -1,0 +1,18 @@
+; File managed by puppet.
+;
+
+include "<%= @cfg_dir %>/named.conf.options";
+logging {
+        channel default_debug {
+                file "data/named.run";
+                severity dynamic;
+        };
+};
+zone "." IN {
+       type hint;
+       file "<%= root_hint %>";
+};
+include "<%= @rfc1912_zones_cfg %>";
+include "<%= @rndc_key_file %>";
+include "<%= @cfg_dir %>/named.conf.local";
+
--- a/templates/named.conf.erb
+++ b/templates/named.conf.erb
@@ -10,7 +10,7 @@ logging {
 };
 zone "." IN {
        type hint;
-       file "<%= root_hint %>";
+       file "<%= @root_hint %>";
 };
 include "<%= @rfc1912_zones_cfg %>";
 include "<%= @rndc_key_file %>";
--- a/manifests/server/params.pp
+++ b/manifests/server/params.pp
@@ -6,6 +6,9 @@ class dns::server::params {
       $cfg_dir            = '/etc/bind'
       $cfg_file           = '/etc/bind/named.conf'
       $data_dir           = '/etc/bind/zones'
+      $root_hint          = "${cfg_dir}/db.root"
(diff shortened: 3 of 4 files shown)


templates/named.conf.erb
```diff
@@ -0,0 +1,18 @@
...
+include "<%= @cfg_dir %>/named.conf.options";
+logging {
+        channel default_debug {
+                file "data/named.run";
+                severity dynamic;
+        };
+};
+zone "." IN {
+       type hint;
+       file "<%= root_hint %>";
```
solarkennedy: use `@root_hint` for puppet 4 compat

solarkennedy:
Thank you for doing this!

I think long term it makes sense for puppet to manage this. I think that it is a natural progression for this module.

This does lead to a potentially breaking change that I would like the other developer to weigh in on.

@ajjahn?

ajjahn:
@solarkennedy Yep, I agree. Probably should ship with a major version bump if it's going to break things. @jearls Thanks.

templates/named.conf.erb
darkfoxprime: Whoops, I missed that one, sorry!  fix committed.

darkfoxprime:
Is this actually a breaking change?  The content is (pretty much) the same as the default Debian config, and the default RedHat config just doesn't work with this module anyway.  I guess it's a breaking change for people who modified the default named.conf by hand?

jearls:
Is this actually a breaking change?  Unless people modified the default named.conf by hand, this replaces the Debian named.conf with equivalent functionality (i think, I don't have a debian box to test on) and makes it actually work on RedHat (where the default named.conf never sees the changes that this module is making)

ajjahn:
@jearls Right, not a breaking change unless the file is manually edited.

@solarkennedy?

solarkennedy:
Right. It's just a new file that this puppet module is editing where it was not before.
I think that is fine, but other people use puppet differently.

Let me do one more code review pass.

manifests/server/config.pp
```diff
@@ -34,6 +34,7 @@
     owner   => $owner,
     group   => $group,
     mode    => '0644',
+    content => template("${module_name}/named.conf.erb"),
```
solarkennedy: is $module_name right here? Not $cfg_dir?

solarkennedy:
Great!

I'll bump the major version so the next time we do a release it will be on a new, major version.

Thanks again @jearls for taking a PR and following up with it!

darkfoxprime:
@solarkennedy go ahead and revert this change.  I thought I had tested it properly, but I guess not.  I'll re-create the changes and submit a new PR at a later time, once I'm sure it really works.

darkfoxprime:
@solarkennedy - to make this work properly in RedHat, I think we're going to need to take over all the default files that named technically provides.  In 6.x, those files are installed under /etc, but in 5.x, they're in `/usrshare/doc/bind-9.3.6/sample/etc/`

By taking over control of all those files, or just incorporating them into the default named.conf, we should be able to work with both Debian and RedHat because we won't be dependent on the files that are installed (or not) by the package.

tedivm:
@darkfoxprime / @jearls -

The issues weren't too hard to resolve, and I took care of them in #119. I'd suggest just merging that back in and forking off the master for corrections.

If the issue is that bind is giving different defaults to RedHat than Debian I'd suggest not trying to take over the /usr/share/doc files (as they can get overwritten during updates) but just putting the required changes into the named.conf file *only* when the system is RedHat. In other words, issue a quick conditional in the template and if it's redhat put the root zone and logging code into named.conf and otherwise leave it along.

PR #113, opened 2015-05-09
Add param to manage packages

jaxim:
Hi - With the environments I work with we need a bit more control of the versions of software installed and can't take the latest versions straight away. This PR is to have a parameter for the packages so that an alternative value such as 'present' can override the default value for the param. I set the default for the new param to 'latest' as that is what the module currently uses.


solarkennedy:
Thanks!

PR #114, opened 2015-05-09
Fix comment syntax in named.conf template

jaxim:
Hi - Was unable to start the named service due to syntax error in named.conf tempate. This PR corrects the syntax on the comments.


solarkennedy:
Thanks!

PR #116, opened 2015-05-15
Updated concat module version

tedivm:
The concat module is on version 2.0.0. The vast majority of changes are internal, meant to speed the module up. On first glance there do not appear to be any changes which would cause this module issue.

### DIFF
--- a/manifests/key.pp
+++ b/manifests/key.pp
@@ -42,7 +42,6 @@ define dns::key {
     owner   => $dns::server::params::owner,
     group   => $dns::server::params::group,
     mode    => '0644',
-    require => Class['concat::setup'],
     notify  => Class['dns::server::service']
   }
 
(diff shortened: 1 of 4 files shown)


tedivm:
I'm having a bit of trouble with this- while it should work, it is not actually creating the needed fragments (in /var/lib/puppet/concat/) or the end files themselves. I am not sure if this is a bug in the DNS module or an issue with the upstream Concat module, but would appreciate anyone having a chance to look it over.

solarkennedy:
Well.. the tests are still passing, which seems to suggest that the catalog works?

Can you reproduce this at the unit test level by adjusting the .fixtures.yml to reflect the new module explicitly? 
https://github.com/puppetlabs/puppetlabs_spec_helper#fixtures-examples

tedivm:
I figured out the issue with the concat module and it was specific to my test installation.

tedivm:
Why was this closed? Are you not planning on supporting the newer concat module?

solarkennedy:
Oh, sorry, I misinterpreted your comment.

solarkennedy:
Thank you. Yes, we do want to keep up to date with concat.

The test pass, so if something is not working correctly, then we are missing some test coverage.

PR #118, opened 2015-05-18
Fix non-existent directory and substitute by a param variable.

pecastro:
CentOS 7 reports ...

May 19 00:20:41 server3 named-checkconf[463]: /etc/named/named.conf.options:2: parsing failed
May 19 00:20:41 server3 named-checkconf[463]: /etc/named/named.conf.options:2: change directory to '/var/cache/bind' failed: file not found

### DIFF
--- a/templates/named.conf.options.erb
+++ b/templates/named.conf.options.erb
@@ -1,5 +1,5 @@
 options {
-	directory "/var/cache/bind";
+	directory "<%= scope.lookupvar('::dns::server::params::data_dir') %>";
 
 	// If there is a firewall between you and nameservers you want
 	// to talk to, you may need to fix the firewall to allow multiple


solarkennedy:
Hmm. This directory needs to be in server/params.pp.

Can you tell me what it *should* be fore centos 7? (and probably all centos)

templates/named.conf.options.erb
```diff
@@ -1,5 +1,5 @@
 options {
-	directory "/var/cache/bind";
+	directory "<%= scope.lookupvar('::dns::server::params::data_dir') %>";
```
solarkennedy: can you use `@data_dir` instead?

This more in-line with the style, and also allows the parameter to be easily overridden.

pecastro: Changing this do data_dir will involve either changing the dns::server::options to accept another param or creating a local variable within the module that is assigned the ::dns::server::params::data_dir variable.

Overriding this is already possible by using hiera and passing in params into the params module.

Your call ...

jearls: For the first point:  The PR includes the change to the dns::server::options class:
  ) {
    $valid_check_names = ['fail', 'warn', 'ignore']
    $cfg_dir = $::dns::server::params::cfg_dir
+  $data_dir = $::dns::server::params::data_dir

    if ! defined(Class['::dns::server']) {
      fail('You must include the ::dns::server base class before using any dns options defined resources')

For your second point:  Overriding this is not currently possible using hiera because it's hardcoded in the template,  unless you're saying that to make this work on RedHat, every user must define their data_dir to be /var/cache/named.

- Johnson

Paulo Edgar Castro wrote:
> In templates/named.conf.options.erb <https://github.com/ajjahn/puppet-dns/pull/118#discussion_r31315448>:
>
>> @@ -1,5 +1,5 @@
>>  options {
>> -	directory "/var/cache/bind";
>> +	directory "<%= scope.lookupvar('::dns::server::params::data_dir') %>";
>
> Changing this do data_dir will involve either changing the dns::server::options to accept another param or creating a local variable within the module that is assigned the ::dns::server::params::data_dir variable.
>
> Overriding this is already possible by using hiera and passing in params into the params module.
>
> Your call ...
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/ajjahn/puppet-dns/pull/118/files#r31315448>.
>

-- 
Johnson Earls
darkfoxprime@gmail.com

darkfoxprime: My apologies! I thought Paulo's comment was on a different PR.  I should make my replies on github not through email.

solarkennedy:
Is @data_dir the right place for this? Seems like it is the equivalent "cache" dir on centos, like /var/cache/named or something?

pecastro:
You guys are ridiculous .... 6fd7c9d84721ad8a757fbf5569f4085f80e67e81

PR #121, opened 2015-05-29
Remove invalid reference to dns::server::options::forwarder

darkfoxprime:
The dns::zone class tried to reference dns::server::options::forwarder as if dns::server::options was a class rather than a defined type.  Removing that check for now.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -24,10 +24,6 @@ define dns::zone (
 
   validate_array($allow_transfer)
   validate_array($allow_forwarder)
-  if $dns::server::options::forwarder and $allow_forwarder {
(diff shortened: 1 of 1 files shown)


darkfoxprime:
Removing the check against having a global forwarder and a per-zone forwarder until I can figure out how to reference an attribute from a resource defined elsewhere.

solarkennedy:
I agree, this is strange and would never work.

Maybe we should give up on dns::server::options being a defined type?

darkfoxprime:
Why _is_ dns::server::options a defined type?  Is there ever a reason to use it more than once?

solarkennedy:
I personally don't think so. I would merge in a PR to make it not so! :)

PR #123, opened 2015-05-29
fix zone template's @allow_transfer check

darkfoxprime:
the zone template was including empty allow_transfer {} blocks because
a 0-length array was still being treated as "true".  this PR changes
the check to verify that @allow_transfer is an array and has >0 size.


solarkennedy:
Thanks!

PR #125, opened 2015-05-29
Fix the `directory` option in named.conf.options

darkfoxprime:
Instead of hardcoding it to point to /var/cache/named (which does not exist on RedHat), use dns::server::params::data_dir instead.

### DIFF
--- a/templates/named.conf.options.erb
+++ b/templates/named.conf.options.erb
@@ -1,5 +1,5 @@
 options {
-	directory "/var/cache/bind";
+	directory "<%= @data_dir %>";
 
 	// If there is a firewall between you and nameservers you want
 	// to talk to, you may need to fix the firewall to allow multiple
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -85,6 +85,7 @@ define dns::server::options (
 ) {
   $valid_check_names = ['fail', 'warn', 'ignore']
   $cfg_dir = $::dns::server::params::cfg_dir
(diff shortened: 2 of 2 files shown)


templates/named.conf.options.erb
```diff
@@ -1,5 +1,5 @@
 options {
-	directory "/var/cache/bind";
```
solarkennedy: This seems fine, but before this being /var/cache/bind, it will change to /etc/bind/zones, is that right? (for debian)

On redhat this will change to "/var/named/" Can you also confirm this is appropriate?


I think that this directory is for the caching resolver right? If so then maybe we should just leave it as /var/cache/bind and have puppet manage it (create it)

darkfoxprime:
No, my understanding is that this is the directory where named expects to find all unqualified filenames.  In RedHat, this defaults to `/var/named`, and zone files in `named.conf` are unqualified.  I guess the Debian default is to put the directory option as `/var/cache/named`, and fully-qualify all zone file names to `/etc/db/zones/`_filename_.

templates/named.conf.options.erb
darkfoxprime: No, my understanding is that this is the directory where named expects to find all unqualified filenames.  In RedHat, this defaults to `/var/named`, and zone files in `named.conf` are unqualified.  I guess the Debian default is to put the directory option as `/var/cache/named`, and fully-qualify all zone file names to `/etc/db/zones/`_filename_.

solarkennedy:
Thanks!

PR #126, opened 2015-06-06
Added initial acceptance test framework

solarkennedy:
@ajjahn This adds the beaker infrastructure so we can accept better PRs and have more confidence that they work.

These tests currently do *not* work on centos because of the silly way we make people have the full filename in the dns::server::options type. (eventually that will become a class and pickup the right filename from params)

I had to remove the logging section because it didn't allow bind to start up! (we are not managing the logging things properly yet)

Additionally I had to order the install->config~>service stuff so thing converged in one run so the test for idempotency passed.


solarkennedy:
Hmm, needs ruby >1.87... that might be hard to support with these newer gems...

ajjahn:
@solarkennedy I haven't looked much at beaker, but I'm all for making it easier to acceptance test on different platforms. Should we keep this in it's own branch until we can work out the breaking changes with certain platforms (CentOS)?

solarkennedy:
@ajjahn Na, I vote we have this on master so that when we fix the centos stuff, we can be sure we really fixed it! (and of course test for regressions)

These tests involve vagrant and junk, they will not run with normal travis runs.

Once I get the gemfile thing sorted, having this framework available can't "hurt" anything.

ajjahn:
@solarkennedy Got it. Merging...

PR #127, opened 2015-06-10
Add Gemfile.lock to .gitignore

darkfoxprime:
I noticed that the `Gemfile.lock` file was removed during a recent update.  Since this file apparently gets recreated when needed, and (from experience) it might be different depending on the version of bundler and other tools in use, I'm suggesting adding it to the `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ pkg
 .vagrant
 /.project
 log/
+Gemfile.lock


solarkennedy:
I did this kinda out of "frustration" trying to get the gems to build. I don't think we should *actually* do this, but I didn't want to pollute what Travis had with whatever gemfile.lock I had that wasn't working.

Can you rebase and instead actually commit *your* Gemfile and see if Travis likes it?

darkfoxprime:
I will try that in the morning.

(wow I really am schizophrenic about my passing identities, aren't I?)

PR #129, opened 2015-06-11
feature TSIG configuration

eumel8:
To exchange DNS zones between two bind installations like master/slave you can sign this transaction with a key. It's more secure as IP-based zone transfer and protect against spoofing (see ftp://ftp.isc.org/isc/bind/9.8.0-P4/doc/arm/Bv9ARM.ch04.html#tsig)

### DIFF
--- a/templates/tsig.erb
+++ b/templates/tsig.erb
@@ -2,8 +2,8 @@ key <%= @keyname %>. {
   algorithm <%= @algorithm %>;
   secret "<%= @secret %>";
 };
-<% if server.class == Array -%>
-<% server.each do |host| -%>
+<% if @server.class == Array -%>
+<% @server.each do |host| -%>
 server <%= host -%> {
   keys { <%= @keyname %>. ; };
 };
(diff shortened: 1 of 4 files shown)


templates/tsig.erb
```diff
@@ -0,0 +1,17 @@
+key <%= @keyname %>. {
+  algorithm <%= @algorithm %>;
+  secret "<%= @secret %>";
+};
+<% if server.class == Array -%>
```
solarkennedy: `@server` ?

solarkennedy:
Can you add a test for this defined type to ensure that the erb syntax is valid?

Maybe do at least two contexts, one with an array of ips for server and one with a single server ip?

eumel8:
@solarkennedy I added some simple checks. Maybe it's enough for the first try.

templates/tsig.erb
solarkennedy: Use `@server` notation in erb variables

eumel8:
overseen

solarkennedy:
Thanks!

PR #130, opened 2015-06-19
Added support to modify service startup

Cicco0:
I needed the possibility to modify the startup service, so I put this in. Now you are able to start bind in IPv4/IPv6 mode or passing some other variables.

In addition I had to fix some issues with concat_basedir to run rpsec tests locally.


manifests/server/startup.pp
```diff
@@ -0,0 +1,90 @@
+# == Define: dns::server::startup
```
solarkennedy: I think we are ready to need this. I don't really like the name "startup" though, it makes me think it handles init script.

How about class: dns::server::defaults_file ?

manifests/server/startup.pp
```diff
@@ -0,0 +1,90 @@
...
+#   Set string "yes", "no" or leave empty. Default: empty
+#
+# === Examples
+#
+#  dns::server::startup { '/etc/default/bind9':
+#    resolvconf => 'no',
+#    options    => '-u bind'
+#  }
+#
+class dns::server::startup (
```
solarkennedy: Is it a define or a class? Should be a class, we only have one o them.

manifests/server/startup.pp
```diff
@@ -0,0 +1,90 @@
...
+# [*keytab_file*]
+#   Must be an absolute path to set up the keytab file
+#
+# [*disable_zone_checking*]
+#   If you set this option to 'yes' then initscript doesn't perform named-checkzone
+#   Set string "yes", "no" or leave empty. Default: empty
+#
+# === Examples
+#
+#  dns::server::startup { '/etc/default/bind9':
```
solarkennedy: I don't want to make the same mistake as the options file. This shouldn't be a defined type that users have to "remember" to add to their manifest. init.pp should include this.

Cicco0: Okay, i will rewrite the code from a class to a define type

manifests/server/startup.pp
Cicco0: It is a class, the documentation is wrong.

manifests/server/startup.pp
Cicco0: How about class dns::server::defaults? The definition of **defaults_file** will be used in the code to set the filename.

solarkennedy: That sounds fine

Cicco0:
Now it is renamed and corrected. Do you repair the merge conflict?

solarkennedy:
It is usually up to the pull requested to fix merge conflicts and bring the PR into a mergable form.

Cicco0:
The merge conflicts are solved :-)

manifests/init.pp
```diff
@@ -6,4 +6,5 @@
   # include dns::install
   # include dns::config
   # include dns::service
+  include dns::server::default
```
solarkennedy: Shouldn't this be in dns::server::install? Seems out of place here

Cicco0:
Ok is fixed and was moved to "manifest/server/install.pp".

manifests/server/default.pp
```diff
@@ -0,0 +1,52 @@
...
+  $rootdir               = undef,
+  $enable_zone_write     = undef,
+  $enable_sdb            = undef,
+  $disable_named_dbus    = undef,
+  $keytab_file           = undef,
+  $disable_zone_checking = undef,
+
+) inherits dns::server::params {
+
+  if ! defined(Class['::dns::server']) {
```
solarkennedy: This is incorrect now, this class should be safe to include no matter what

manifests/server/install.pp
```diff
@@ -1,5 +1,7 @@
 # == Class dns::server
 #
+include dns::server::default
```
solarkennedy: This is not correct, you can't put an include outside of a class definition

solarkennedy: The default class should be chained after the package.
The package should be installed first, then the default class will replace the stock defaults file in that order.

Cicco0: I would now solve it like this in the file **server.pp**:
```
class dns::server {
  class { 'dns::server::install': } ->
  class { 'dns::server::config': } ->
  class { 'dns::server::default': } ~>
  class { 'dns::server::service': }
}
```
Do you agree with this changement?

jearls: You can do this in the puppet 4 / future parser, does this not work in the traditional parser?

Kyle Anderson wrote:
> In manifests/server/install.pp <https://github.com/ajjahn/puppet-dns/pull/130#discussion_r33468120>:
>
>> @@ -1,5 +1,7 @@
>>  # == Class dns::server
>>  #
>> +include dns::server::default
>
> This is not correct, you can't put an include outside of a class definition
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/ajjahn/puppet-dns/pull/130/files#r33468120>.
>

-- 
Johnson Earls
darkfoxprime@gmail.com

solarkennedy: It does, but you shouldn't. All resources need to be contained in some sort of class, not just setting out there for the parser to suck up as it reads the file.

solarkennedy: @Cicco0 yea I think that would work, but I think I actually prefer it to be in the config class, as it is a config file?

Cicco0: @solarkennedy I've included it:
`include dns::server::default`

Now everything should be fine. If so, you need a rebase?

solarkennedy: instlal.pp:3 still needs to be removed. It is a include line outside a class definition and will be included by simply parsing this file, which is surprising and impossible to test.

Cicco0: Was removed and tests were positive.

solarkennedy:
Good.

We have an acceptance test suite using beaker that is now failing. We don't run these on Travis as they are too intensive.

You can run these tests with `bundle exec rake beaker` per the contributing doc:
https://github.com/ajjahn/puppet-dns/blob/master/CONTRIBUTING.md#integration-tests

I ran them here and here are the failures:
http://pastebin.com/JJWqzBqA

Cicco0:
I'll try it. Apparently I do not get it to work "out of the box".

Cicco0:
Now, after several more days, the problem could be solved. All tests were carried out and I get no error.

solarkennedy:
Good. Can you pastebin the output?

Cicco0:
Sure:
 - bundle exec rake test
 - - http://pastebin.com/Hw8JGfVk
- bundle exec rake spec
 - - http://pastebin.com/5SZYS1Zq
. bundle exec rake beaker
 - - http://pastebin.com/LQbYJ0D8

solarkennedy:
Thank you!

PR #131, opened 2015-06-22
params.pp: excluded dnssec-tools from $necessary_package for debian 8

Gril258:
hi i have excluded dnssec-tools for debian 8, there is no package in official repository for jessie.


solarkennedy:
The tests are failing for a different new reason...
But thank you.

Out of curiosity, how do you `'dnssec-signzone` on that distro?

PR #132, opened 2015-07-24
Move stats Channels

acidprime:
Prior to this commit the stats channels was inside options which caused bind to not start
Moving this allowed the daemon to start and verified it was a working configuration as well.

```
2015-07-23T22:56:15.854601-04:00 hank named[4311]: /etc/bind/named.conf.options:29: unknown option 'statistics-channels'
```


solarkennedy:
@acidprime we have an acceptance test to catch this kind of error:
https://github.com/ajjahn/puppet-dns/blob/master/CONTRIBUTING.md#integration-tests

Can you write or example the test suite to catch this?
(you do have to run these tests manually yourself)

Please try to write a failing test first (I think a test that includes the stats channel params), then add in your fix to prove that it really does fix it.

PR #133, opened 2015-07-28
Allow the dns::zone::slave_masters parameter to be an array

jearls:
This just changes the `zone.erb` template to allow the `@slave_masters` parameter to be an array, in which case it gets joined by `";"` semicolons; if it's not an array, it's interpolated directly, just like the previous behavior.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -1,5 +1,136 @@
 # == Define dns::zone
 #
+# `dns::zone` defines a DNS zone and creates both the zone entry
+# in the `named.conf` files and the standardized zone file header
+# to which the zone records will be added.
+#
+# === BIND Time Values
+#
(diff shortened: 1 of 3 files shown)


jearls:
note - i am working on adding spec tests to this; please hold :)

solarkennedy:
This is good. While we have not been doing this regularly, can you add a docstring to `manifests/zone.pp` documenting this parameter?

Something like

```
# == Parameters
#
# [*slave_masters*]
# .....
#
```

jearls:
I'll add in documentation for all the parameters this evening.

jearls:
parameters documented

solarkennedy:
Whoa! Awesome!!
Thank you!

ajjahn:
@jearls wins all the points.

PR #134, opened 2015-08-03
Make "listen-on-v6" a configurable option

djm256:
Previously, the "listen-on-v6" option was hard-coded to the value "any" which limited IPv6 configurability.

This commit adds dns::server::options::listen_on_ipv6, allowing an array of values to be provided for the config option.

### DIFF
--- a/spec/defines/dns__server__options_spec.rb
+++ b/spec/defines/dns__server__options_spec.rb
@@ -70,6 +70,26 @@ describe 'dns::server::options', :type => :define do
     it { should raise_error(Puppet::Error, /is not an Array/) }
   end
 
+  context 'when passing valid array to listen_on_ipv6' do
+    let :params do
+      { :listen_on_ipv6 => [ '2001:db8:1::1', '2001:db8:2::/124' ] }
+    end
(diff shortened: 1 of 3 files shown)


djm256:
Hi.  Here's the output of a successful `bundle exec rake beaker` run: https://gist.github.com/djm256/606dda198806d6a91378

spec/defines/dns__server__options_spec.rb
```diff
@@ -70,6 +70,26 @@
...
 
+  context 'when passing valid array to listen_on_ipv6' do
+    let :params do
+      { :listen_on_ipv6 => [ '2001:db8:1::1', '2001:db8:2::/124' ] }
+    end
+    it { should contain_file('/etc/bind/named.conf.options').with_content(/2001:db8:1::1;$/)  }
+    it { should contain_file('/etc/bind/named.conf.options').with_content(/2001:db8:2::\/124;$/)  }
+  end
+
+  context 'when passing a string to listen_on_ipv6' do
```
solarkennedy: I don't think you realy have to double check the validate_array function, it will be fine

solarkennedy:
Thanks!

djm256:
Glad I could contribute.  Thanks for a great module :+1:

PR #135, opened 2015-09-03
fixes for puppet future parser support

timogoebel:
When using puppet future parser, this module does not work when e.g. resolvconf is not set (read: undef).
```
Error: The resolvconf value is not type of a string yes / no. on node example.xyz
```

This PR fixes this. Maybe you should have travis test Puppet 4 and future parser as well. But I think, this is out of scope of this PR.

### DIFF
--- a/manifests/server/default.pp
+++ b/manifests/server/default.pp
@@ -18,11 +18,11 @@ class dns::server::default (
 
   validate_absolute_path( $default_file )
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
Thanks! I don't use Puppet4/Future yet in production, but we would certainly accept a PR to add support for it!

PR #136, opened 2015-09-03
fix variable access preference with @preference

timogoebel:
This PR fixes the warning:

```
Warning: Variable access via 'preference' is deprecated. Use '@preference' instead. template[/path/to/puppet/modules/dns/templates/zone_record.erb]:7
   (at /usr/lib/ruby/vendor_ruby/puppet/parser/templatewrapper.rb:76:in `method_missing')
```


solarkennedy:
Thanks!

PR #137, opened 2015-09-30
Add `all_ptr` parameter to `dns::record::a` record type

jearls:
This adds the `all_ptr=>`_[true|false]_ parameter to the `dns::record::a` type and sets up a private defined type that allows the `dns::record::a` type to create multiple `dns::record::ptr` resources for a records with multiple `data` values.

This also creates a set of rspec tests for the `dns::record::a` defined type.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -7,3 +7,85 @@ class dns {
   # include dns::config
   # include dns::service
 }
+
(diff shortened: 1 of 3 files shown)


jearls:
i'm going to cancel and re-create based on a current master, so the "merge" commit isn't stuck in there.

PR #140, opened 2015-10-26
Test NS records, provide example for README

roderickm:
Build upon the work contributed in https://github.com/ajjahn/puppet-dns/pull/53 and https://github.com/ajjahn/puppet-dns/pull/96 by adding tests and a README example.


solarkennedy:
Thanks!

Issue #142, opened 2015-12-03
Concat dependency causing builds to fail


ajjahn:
At first I thought this was due to the version bump commit, but it looks like even old builds that once passed are now failing.

@solarkennedy Any ideas?

PR #143, opened 2015-12-04
Add support for delegation-only zone types.

evidex:
Adds support for delegation-only zone types as specified on Pg. 105 of [Bind 9.10 Manual](https://www.isc.org/wp-content/uploads/2014/01/Bv910ARM.pdf). 

Added rspec test to ensure that a `dns::zone` with type `delegation-only` contains no other statements.


evidex:
I think the failing builds are related to #142 ?

solarkennedy:
Thanks!

PR #145, opened 2015-12-04
[WIP] Fix fixtures

### LINKED ISSUE #142: Concat dependency causing builds to fail

solarkennedy:
Yea I think we would be better served using released versions of modules instead of HEAD on them.
I'll try to set that which should close #142


solarkennedy:
Thanks!

PR #146, opened 2015-12-04
Control whether DNS-SEC support is enabled/disabled

evidex:
Add boolean option to `dns::server::options` to control whether DNS-SEC support is enabled/disabled. Also including tests & docs for same.

This option is default to false on RedHat 5 due to lack of DNS-SEC support, and true on all other platforms.


solarkennedy:
This looks correct to me. Thanks!

PR #153, opened 2016-01-26
Added Package require

mooreandrew:
When trying to integrate this module with my puppet environment, it kept claiming that the named user wasn't present. This was because the package hadn't yet been installed.

Error: Could not set 'present' on ensure: Could not find user named at 46:/vagrant/puppet-control/modules/dns/manifests/server/default.pp
Error: Could not set 'present' on ensure: Could not find user named at 46:/vagrant/puppet-control/modules/dns/manifests/server/default.pp
Wrapped exception:
Could not find user named
Error: /Stage[main]/Dns::Server::Default/File[/etc/sysconfig/named]/ensure: change from absent to present failed: Could not set 'present' on ensure: Could not find user named at 46:/vagrant/puppet-control/modules/dns/manifests/server/default.pp

This change makes sure the the file doesn't get created before the package (and user) is there.

### DIFF
--- a/manifests/server/default.pp
+++ b/manifests/server/default.pp
@@ -43,7 +43,7 @@ class dns::server::default (
     mode    => '0644',
     content => template("${module_name}/${default_template}"),
     notify  => Class['dns::server::service'],
-    require => Package[$necessary_packages]
+    require => Package[$::dns::server::params::necessary_packages]
   }
 
 }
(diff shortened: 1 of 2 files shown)


ajjahn:
Thank you. Could you tack on a spec that tests this behavior?

mooreandrew:
Added Spec Tests

solarkennedy:
Thanks!

PR #158, opened 2016-04-30
Removed 'ensure' setting from concat::fragment statements

Loewe88:
The ensure parameter is deprecated for concat::fragment since [v2.0.0 of puppetlabs-concat](https://github.com/puppetlabs/puppetlabs-concat#ensure-1)

On systems with puppetlabs-concat v2.0.0 the ensure parameter is ignored for concat::fragment and it works just fine, so I removed the ensure settings.


solarkennedy:
Thankyou. Looks like the tests need to be updated but that isn't related to this change.

PR #164, opened 2016-05-16
Add `serial_style` parameter to dns::zone

jearls:
This adds an option to dns::zone to generate the serial number using
the default 'timestamp' style, where the serial number is the unix
timestamp at the time of serial generation, or a new 'date' style
that is the traditional YYYYMMDDXX serial number of zone files managed
by hand.

The serial number generation itself is simple, but the command to apply
the serial number to the zone file is now much more complex, requiring an
examination of the existing zone file to determine the current serial number
so that the new serial number can be incremented above the current if needed.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -85,6 +85,11 @@
 #   to it.
 #   Defaults to `false`.
 #
+# [*serial_style*]
+#   The style of the serial number.  This should be either `date`
+#   for the YYYYMMDDXX style of serial number, or `timestamp` for
+#   a straight unix timestamp serial number.  Defaults to `timestamp`.
+#
 # [*slave_masters*]
 #   If *zone_type* is set to `slave`, this holds an array or string
 #   containing the IP addresses of the DNS servers from which this slave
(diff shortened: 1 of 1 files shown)


solarkennedy:
I'm pretty reluctant to merge this due to the complexity.
What is the particular use case here? These zones are machine-generated, is it just so a human can read the zone file? If so what about a humanized date as a comment in the zone file?

jearls:
the use case is for someone (like me) who is trying to convert from
manually managed zone files to puppet controlled files.  the traditional
YYYYMMDDXX serial format of manual zone files is unfortunately a greater
number than the UNIX timestamp used to create the serial number for zones
managed by this module.  this will cause problems when trying to convert,
since i have secondary name servers over which i have no control, and those
servers will stop updating if the serial number moves backwards.

jearls:
Well, I'll withdraw this, since I've figured out a [painful] way to work around the issue (I had to cancel my secondary service, change the serial numbers to something that would be lower than those generated by this module, re-establish the secondary service, and then deploy this module).

Issue #170, opened 2016-08-15
TXT record types should properly format the data value


jearls:
DNS TXT records have specific formatting requirements.
1. per [RFC1035 section 3.3.14](https://tools.ietf.org/html/rfc1035#section-3.3.14), they are composed of one *or more* character strings.
2. per [RFC1035 section 3.3](https://tools.ietf.org/html/rfc1035#section-3.3), each character string can be up to 255 characters (section 3.3 indicates 256 octets, but includes the length octet as part of that count).
3. per common usage, if a TXT record consists of multiple character strings, the application processing that record concatenates the character strings together with no intervening characters.
4. per [RFC1035 section 5](https://tools.ietf.org/html/rfc1035#section-5), if a character string includes spaces, it must be enclosed in double-quotes (`"`).  Within the double-quotes, any character can occur except another double-quote; double-quotes must be escaped by backslashes (`\`) (and by implication, backslashes must also be escaped by backslashes).

Based on these requirements, the `dns::record::txt` record type should:
1. split the `$data` value up into 255 (at most) character chunks.
2. escape all double-quotes and backslashes within each chunk by inserting a backslash before the escaped character.
3. surround each chunk with double-quotes.
4. separate the chunks by a space.

For example:

    dns::record::txt { 'txt1.example.com':
        host => "txt1' ,
        zone => 'example.com' ,
        data => 'test "number one"' ,
    }

should result in the following `dns::record` resource:

    dns::record { 'txt1.example.com,TXT,example.com':
        zone   => 'example.com' ,
        host   => 'txt1' ,
        ttl    => '' ,
        record => 'TXT' ,
        data   => '"test \"number one\""'
    }

And a text record with a very long data value ('this is a very _(x60)_ long test') should result in something like...

    dns::record { 'txt2.example.com,TXT,example.com':
        zone   => 'example.com' ,
        host   => 'txt2' ,
        ttl    => '' ,
        record => 'TXT' ,
        data   => '"this is a very very very…very " "very very very very very very very very very very very long test"'
    }

I will submit a PR for this shortly.

darkfoxprime:
This issue can be closed with the merging of #171 .

PR #171, opened 2016-08-15
issue #170: Produce proper DNS quoted strings for TXT and SPF records

jearls:
#170 This change modifies the `zone_record.erb` template so that it
processes the data values for TXT and SPF records to produce proper
DNS quoted strings (of no more than 255 characters, with embedded
quotes and backslashes escaped with backslashes).

It also adds SPEC tests for the TXT record type.

### DIFF
--- a/spec/defines/dns__record__txt_spec.rb
+++ b/spec/defines/dns__record__txt_spec.rb
@@ -14,7 +14,7 @@ describe 'dns::record::txt', :type => :define do
     it { should contain_concat__fragment('db.example.com.txttest,TXT,example.com.record').with_content(/^txttest\s+IN\s+TXT\s+"testing"$/) }
   end
 
-  context 'passing a string that includes a quote should result in the quote being escaped' do
+  context 'passing a string that includes a quote character should result in the dns module escaping the quote' do
     let :params do {
         :host => 'txttest',
         :zone => 'example.com',
@@ -24,7 +24,7 @@ describe 'dns::record::txt', :type => :define do
     it { should contain_concat__fragment('db.example.com.txttest,TXT,example.com.record').with_content(/^txttest\s+IN\s+TXT\s+"this is a \\"test\\""$/) }
   end
 
-  context 'passing a long string should result in multiple quoted strings' do
+  context 'passing a long string should result in the dns module splitting that string into multiple quoted strings' do
     let :params do {
         :host => 'txttest',
         :zone => 'example.com',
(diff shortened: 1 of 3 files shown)


templates/zone_record.erb
```diff
@@ -1,8 +1,14 @@
 <%
-delimiter = %w[TXT SPF].include?(@record.upcase) ? '"' : ''
 data_a = @data.is_a?(String) ? @data.lines.to_a : @data
 
 data_a.each do |data_item|
+if %w[TXT SPF].include?(@record.upcase)
+  # for TXT and SPF records, transform the data item into a valid text record value:
+  # use String.unpack to split the text value into 255 character chunks;
+  # then use Array.collect to turn each chunk into a valid quoted character string;
+  # finally join the quoted character strings together, separated by a space.
```
solarkennedy: You did a great job documenting why we do this in the issue, can you put the abridged version here? (at least link to the RFC)

spec/defines/dns__record__txt_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  context 'passing a long string should result in multiple quoted strings' do
+    let :params do {
+        :host => 'txttest',
+        :zone => 'example.com',
+        :data => 'this is a ' + 'very '*60 + 'long test',
+    } end
+    it { should_not raise_error }
+    it { should contain_concat__fragment('db.example.com.txttest,TXT,example.com.record').with_content(/^txttest\s+IN\s+TXT\s+"this is a very.*" ".*very long test"$/) }
```
solarkennedy: Can you add a note to the TXT/SPF puppet defines to indicate that this escaping is done for you so the user doesn't think they have to do it?

solarkennedy:
Still just want a docstring so it is noted "somewhere" that a user doesn't have to do their own escaping.

Also @danzilio @ajjahn, @jearls is the \#3 contributor and the last pull requests have all only needed minor adjustments. I recommend giving them collaborator access. (so a least one less fork)

jearls:
I guess I don't understand what you are asking for.  What do you mean by a docstring?  Do you mean you want the comment block at the top of `manifests/record/txt.pp` filled out?

solarkennedy:
Yes. I would link you to the "puppet doc" format to that but they've removed in favor of the "puppet strings" stuff which isn't fully baked yet.

jearls:
Markdown source code for last version of Puppet style guide that included the Puppet Doc section:  https://raw.githubusercontent.com/puppetlabs/puppet-docs/462e3350c548d2b0ee9c91d2a68aabb9b6d2c87e/source/guides/style_guide.markdown  -  go to the end for section 13 on Puppet Doc

solarkennedy:
Thanks!

PR #175, opened 2016-09-06
Ignore /*.lock files

sspreitzer:


sspreitzer:
Who needs lock files in git? :-)

solarkennedy:
Well, depending on which rubyist you ask, Gemfile.lock files should be committed because they represent a snapshot of all the dependencies and allows reproducible builds.

Granted, this is a puppet module and we blow away gemfile.lock anyway at travis time.

solarkennedy:
Looks like puppetlabs does this on all their modules too: 
https://github.com/puppetlabs/modulesync_configs/blob/bac142b47f8a6871eef3ee2ca03d666012d0e9f9/moduleroot/.gitignore#L3

sspreitzer:
Ah nice, thank you!

PR #176, opened 2016-09-07
Bump version for new feature

sspreitzer:


solarkennedy:
@ajjahn This is best done through the blacksmith tool for now.

Also would you be ok if I setup travis-ci based releases so any contributor could publish? It would require using travis's password encryption for your forge password. (afaik you can't auth multiple users to the puppet forge for a single module)

sspreitzer:
Interesting approach, please mind security. Pull request travis builds could expose the password and lead to account hijacking..

solarkennedy:
Should be fine. It is standard practice on all puppetlabs stuff:
https://github.com/voxpupuli/puppet-corosync/blob/fa651a72228a3344b324bce8d360e59fa7c08105/.travis.yml#L76

sspreitzer:
Don't know much about travis security features, my concern is about someone "echo"ing the encrypted value to the build output and then hijacking the account. If this is impossible, then it's nice!

ajjahn:
@solarkennedy I'm for it. If you want to, you can set it up and let me know what you need me to do to get you the encrypted password.

@sspreitzer @solarkennedy I'm closing this issue and opening one for the travis-ci releases. Until that's up and running I'll handle bumping the version upon releasing.

PR #178, opened 2016-09-08
Automated Travis Releases

### LINKED ISSUE #177: Setup Travis CI Releases to Forge
Suggested in #176.

solarkennedy:
Closes #177. Primary reviewer: @ajjahn 

I've tested this procedure on some other personal modules. It does work.

See the instructions for how to add an encrypted password and see if the docs on how to make a release are sane. (It is mostly splitting up the normal blacksmith steps and letting travis do the final "module:push" step)

Our travis matrix is pretty complex, and you only want one one travis target to deploy. Instead of a matrix with excludes we could do an include matrix:
https://github.com/voxpupuli/puppet-corosync/blob/master/.travis.yml#L17-L65
But in general we could probably reduce how complex the matrix is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ rvm:
   - 2.1
 
 env:
-  - PUPPET_VERSION="~> 2.7.0"
+  - PUPPET_VERSION="~> 2.7.0" DEPLOY_TO_FORGE=yes
   - PUPPET_VERSION="~> 3.1.0"
   - PUPPET_VERSION="~> 3.2.0"
   - PUPPET_VERSION="~> 3.3.0"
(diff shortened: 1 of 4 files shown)


sspreitzer:
Question @solarkennedy; Do you have a good link for me to get familiar with blacksmith?

solarkennedy:
https://github.com/voxpupuli/puppet-blacksmith
https://kartar.net/2014/10/using-puppet-blacksmith/

No, the github_changelog_generator step will make a changelog:
https://github.com/ajjahn/puppet-dns/pull/178/files#diff-04c6e90faac2675aa89e2176d2eec7d8R199
It is optional but nice to have. We could put that in the Gemfile too if you want, but it requires an API key. (I have it globally installed because I use it on all my github projects, ruby or otherwise)

PR #179, opened 2016-09-15
Disable case sensitivity

MemberIT:
List of changes:
- Disable sensitivity case in defines dns::record::ptr::by_ip and dns::zone;
- Corrected work of defines dns::record::ptr::by_ip and dns::zone;
- Added tests for twin domain zones.

### DIFF
--- a/manifests/record/ns.pp
+++ b/manifests/record/ns.pp
@@ -22,7 +22,7 @@ define dns::record::ns (
   if $check_parameter_reverse != 'reverse' and $check_parameter_reverse != true and $zone =~ /^[0-9\.]+$/ {
     fail("Define[dns::record::ns]: NS zone ${zone} must be a valid domain name.")
   }
-  if ( $check_parameter_reverse == 'reverse' or $check_parameter_reverse == true ) and $zone !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1})$/ {
+  if ( $check_parameter_reverse == 'reverse' or $check_parameter_reverse == true ) and "${zone}." !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1,3})$/ {
     fail("Define[dns::record::ns]: NS zone ${zone} PTR, when parameter 'reverse' of zone '${zone}' is not false, must be included octets with separator dot!")
   }
   # Highest label (top-level domain) must be alphabetic
--- a/manifests/record/ns.pp
+++ b/manifests/record/ns.pp
@@ -14,11 +14,19 @@ define dns::record::ns (
   validate_string($data)
   validate_string($host)
 
-  if !is_domain_name($zone) or $zone =~ /^[0-9\.]+$/ {
+  # Check corrected ns zone
+  $check_parameter_reverse = getparam(Dns::Zone[$zone], 'reverse')
+  if $check_parameter_reverse != 'reverse' and $check_parameter_reverse != true and !is_domain_name($zone) {
     fail("Define[dns::record::ns]: NS zone ${zone} must be a valid domain name.")
   }
+  if $check_parameter_reverse != 'reverse' and $check_parameter_reverse != true and $zone =~ /^[0-9\.]+$/ {
+    fail("Define[dns::record::ns]: NS zone ${zone} must be a valid domain name.")
+  }
+  if ( $check_parameter_reverse == 'reverse' or $check_parameter_reverse == true ) and $zone !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1})$/ {
+    fail("Define[dns::record::ns]: NS zone ${zone} PTR, when parameter 'reverse' of zone '${zone}' is not false, must be included octets with separator dot!")
+  }
   # Highest label (top-level domain) must be alphabetic
-  if $zone =~ /\./ and $zone !~ /\.[A-Za-z]+$/ {
+  if $check_parameter_reverse != 'reverse' and $check_parameter_reverse != true and $zone =~ /\./ and $zone !~ /\.[A-Za-z]+$/ {
     fail("Define[dns::record::ns]: NS zone ${zone} must be a valid domain name.")
   }
   # RR data must be a valid hostname, not entirely numeric values
--- a/manifests/record/ptr/by_ip.pp
+++ b/manifests/record/ptr/by_ip.pp
@@ -139,7 +139,7 @@ define dns::record::ptr::by_ip (
     $reverse_zone = $reverse_class_a_zone
     $octet = $class_a_host
   } else {
-    notify { "PTR record for ip-address '${ip}' not creates! Not found declareted define Dns::Zone of 'A', 'B' or 'C' network classes for ip-address '${ip}'!": }
+    notify { "PTR record for IP address '${ip}' not created: No `dns::zone` resource declared for the class A, B, or C subnet of this IP address.": }
   }
 
   if $reverse_zone and $octet {
--- a/manifests/record/ptr/by_ip.pp
+++ b/manifests/record/ptr/by_ip.pp
@@ -34,9 +34,9 @@
 # turns into:
 #
 # @example
-#     dns::record::ptr { '42.128.168.192.IN-ADDR.ARPA':
+#     dns::record::ptr { '42.128.168.192.in-addr.arpa':
 #       host => '42' ,
-#       zone => '128.168.192.IN-ADDR.ARPA' ,
+#       zone => '128.168.192.in-addr.arpa' ,
 #       data => 'server.example.com' ,
 #     }
 #
@@ -50,21 +50,21 @@
 # turns into:
 #
 # @example
-#     dns::record::ptr { '68.128.168.192.IN-ADDR.ARPA':
+#     dns::record::ptr { '68.128.168.192.in-addr.arpa':
 #       host => '68' ,
-#       zone => '128.168.192.IN-ADDR.ARPA' ,
+#       zone => '128.168.192.in-addr.arpa' ,
 #       data => 'multihomed-server.example.com' ,
 #
 # @example
-#     dns::record::ptr { '69.128.168.192.IN-ADDR.ARPA':
+#     dns::record::ptr { '69.128.168.192.in-addr.arpa':
 #       host => '69' ,
-#       zone => '128.168.192.IN-ADDR.ARPA' ,
+#       zone => '128.168.192.in-addr.arpa' ,
 #       data => 'multihomed-server.example.com' ,
 #
 # @example
-#     dns::record::ptr { '70.128.168.192.IN-ADDR.ARPA':
+#     dns::record::ptr { '70.128.168.192.in-addr.arpa':
 #       host => '70' ,
-#       zone => '128.168.192.IN-ADDR.ARPA' ,
+#       zone => '128.168.192.in-addr.arpa' ,
 #       data => 'multihomed-server.example.com' ,
 #     }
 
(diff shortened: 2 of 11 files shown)


solarkennedy:
This is a lot, and I don't think the case insensitivity is worth the complexity.
I think it is totally fine to just downcase things and let the normal puppet duplicate resource detector find duplicates when users do that.

I can't really distinguish the difference between all the multi-zone work and the downcase work. Can you resubmit with fewer changes at a time?

jearls:
If you're going to be redoing the checks for valid `zone` parameters, as a suggestion, why not move all of those checks out of the `dns::record::*` defines and directly into the `dns::record` define so that all the record types will get the same checks?

jearls:
@solarkennedy in order to simplify both the validity check for zone names and the automatic generation of reverse pointer records, I was going to submit a PR that would modify `dns::zone` to generate a new `dns::zone` record when `reverse` was set - so `dns::zone { '192.168.128': reverse => reverse }` would generate a new `dns::zone { '128.168.192.in-addr

manifests/record/ns.pp
```diff
@@ -14,11 +14,19 @@
...
-  if !is_domain_name($zone) or $zone =~ /^[0-9\.]+$/ {
+  # Check corrected ns zone
+  $check_parameter_reverse = getparam(Dns::Zone[$zone], 'reverse')
+  if $check_parameter_reverse != 'reverse' and $check_parameter_reverse != true and !is_domain_name($zone) {
     fail("Define[dns::record::ns]: NS zone ${zone} must be a valid domain name.")
   }
+  if $check_parameter_reverse != 'reverse' and $check_parameter_reverse != true and $zone =~ /^[0-9\.]+$/ {
+    fail("Define[dns::record::ns]: NS zone ${zone} must be a valid domain name.")
+  }
+  if ( $check_parameter_reverse == 'reverse' or $check_parameter_reverse == true ) and $zone !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1})$/ {
```
jearls: This regex can be simplified to ... `"${zone}." !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1,3})$?` - by adding a `.` to the end of `${zone}`, you don't need all the special cases; you just need to check for 1 to 3 of a valid octet followed by `.`

manifests/zone.pp
```diff
@@ -182,10 +187,31 @@
     fail("The zone_notify must be ${valid_zone_notify}")
   }
 
-  $zone = $reverse ? {
-    'reverse' => join(reverse(split("arpa.in-addr.${name}", '\.')), '.'),
-    true      => "${name}.in-addr.arpa",
-    default   => $name
+  case $reverse {
+    'reverse': {
+      if $name !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1})$/ {
```
jearls: See my comment earlier about regex simplification.

manifests/zone.pp
```diff
@@ -182,10 +187,31 @@
...
+        fail("Name zone with parameter reverse equal 'reverse' must be included only one, two or three first octets! Example: '192.168'!")
+      }
+      $zone = join(reverse(split("arpa.in-addr.${name}", '\.')), '.')
+    }
+    true: {
+      if $name !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1})$/ {
```
jearls: See my comment earlier about regex simplification.

manifests/zone.pp
```diff
@@ -182,10 +187,31 @@
...
+      $zone = "${name}.in-addr.arpa"
+    }
+    default: {
+      if !is_domain_name($name) {
+        fail("Name zone with parameter reverse equal false must be valid domain name! Example: 'exampe.com'!")
+      }
+      $zone = $name
+      if $name =~ /\.in-addr\.arpa$/ {
+        $name_zone_reverse_true    = regsubst($name, '\.in-addr\.arpa$', '')
+        if $name_zone_reverse_true !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1})$/ {
```
jearls: See my comment earlier about regex simplification.

jearls:
I will be submitting a PR for an alternate `dns::zone` solution to help simplify a lot of these checks and help standardize other parts of the code.  The main changes will be:

1.  with `reverse` set to `true` or `reverse`, instead of modifying things in-line, a new `dns::zone` resource will be created with the standardized zone name.
2.  to support the above without breaking the filenames of existing installations, a new optional `filename` parameter will be added that defaults to `$name`.  with `reverse`, the new generated `dns::zone` resource will include the same `filename` parameter as the original resource.
3.  the `filename` parameter will allow the `dns::record` type to correctly identify the `concat` resource to which the new record should be added.

_(Sorry for the earlier broken comment - my system freaked out and I hadn't realized it had submitted)_

jearls:
I'm working on my re-worked `dns::zone` implementation, and realized something:  Even with the new `dns::zone` that allows easy verification that a zone exists and easier access to the filename in which its defined, doing any sort of record validation against that is still a breaking change - there is no requirement now that records must be defined before their DNS zone is defined.

@solarkennedy , @ajjahn , would you want to introduce the requirement that DNS zones be defined before any records are added to that zone, in order to improve the validation of DNS records?

manifests/record/ptr/by_ip.pp
```diff
@@ -76,36 +76,78 @@
...
+    $reverse_zone = $reverse_true_class_a_zone
+    $octet = $class_a_host
+  # Conditions (A.B.C|A.B|A) and parameter reverse is 'reverse'
+  } elsif defined(Dns::Zone[$reverse_class_a_zone]) and getparam(Dns::Zone[$reverse_class_a_zone], 'reverse') == 'reverse' {
+    $reverse_zone = $reverse_class_a_zone
+    $octet = $class_a_host
   } else {
-    $reverse_zone = $class_c_zone
-    $octet = $class_c_host
+    notify { "PTR record for ip-address '${ip}' not creates! Not found declareted define Dns::Zone of 'A', 'B' or 'C' network classes for ip-address '${ip}'!": }
```
MemberIT: Thank you. My english is not good. Sorry for this now and in future.
Fixed.

manifests/zone.pp
```diff
@@ -182,10 +187,31 @@
...
+    }
+    true: {
+      if $name !~ /^((25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]\.|2[0-4][0-9]\.|[01]?[0-9][0-9]?\.){1}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1})$/ {
+        fail("Name zone with parameter reverse equal true must be included only reversed one, two or three first octets! Example: '168.192'!")
+      }
+      $zone = "${name}.in-addr.arpa"
+    }
+    default: {
+      if !is_domain_name($name) {
+        fail("Name zone with parameter reverse equal false must be valid domain name! Example: 'exampe.com'!")
```
MemberIT: Fixed.

manifests/zone.pp
MemberIT: Ok

manifests/record/ns.pp
MemberIT: Thanks. Concatenation dots in end of name zone not needed for correct work.

manifests/record/ptr/by_ip.pp
```diff
@@ -76,36 +76,78 @@
...
   # split the IP address up into three host/zone pairs based on class A, B, or C splits:
   # For IP address A.B.C.D,
-  # class C => host D / zone C.B.A.IN-ADDR.ARPA
-  # class B => host D.C / zone B.A.IN-ADDR.ARPA
-  # class A => host D.C.B / zone A.IN-ADDR.ARPA
-  $class_c_zone = inline_template('<%= @ip.split(".").reverse()[1..3].join(".") %>.IN-ADDR.ARPA')
+  # class C => host D / zone C.B.A.in-addr.arpa
+  # class B => host D.C / zone B.A.in-addr.arpa
+  # class A => host D.C.B / zone A.in-addr.arpa
+  $class_c_zone = inline_template('<%= @ip.split(".").reverse()[1..3].join(".") %>.in-addr.arpa')
```
MemberIT: You can rename zone name from uppercase|camelcase to lowercase and your zone will be corrected.
I think is not good way use different registers names for one zone.
If we will use names only lowercase, we can guarantee for one zone exists one file of zone. We can't declare same zones:
```
  dns::zone { 'example.com':
    soa         => 'ns1.example.com',
    soa_email   => 'admin.example.com',
    nameservers => ['ns1']
  }
  dns::zone { 'eXample.Com':
    soa         => 'ns1.example.com',
    soa_email   => 'admin.example.com',
    nameservers => ['ns1']
  }
```
For reverse twin zones:
```
dns::zone { "168.192.in-addr.arpa":
  reverse => false,
}
dns::zone { "192.168":
  reverse => "reverse",
}
```
we have puppet error: "Duplicate declaration: Exec[bump-168.192.in-addr.arpa-serial] is already declared in file..." 
https://github.com/ajjahn/puppet-dns/pull/179/files#diff-2e38170d1d92d7eafada3bbc9bf83100R385

manifests/zone.pp
MemberIT: Ok

manifests/zone.pp
MemberIT: Ok

manifests/record/ns.pp
MemberIT: Sorry concatenation is needed.

MemberIT:
> I think it is totally fine to just downcase things and let the normal puppet duplicate resource detector find duplicates when users do that.

It is not correct working now, if i use currect master version (commit:d6407bb4aad6876ff653ccb783bf8d417103198b)

Versions:
```
# puppet -V
4.5.3
# hiera -v
3.2.0
# facter -v
3.3.0 (commit fc9282a2d426f7c8a742fd5b23c4cb06852eee20)

```
Hiera data:
```
dns::zone:
  example.com:
    ensure: present
    soa: 'ns1.example.com'
    soa_email: 'admin.example.com'
    nameservers:
      - 'ns1'
  Example.com:
    ensure: present
    soa: 'ns1.example.com'
    soa_email: 'admin.example.com'
    nameservers:
      - 'ns1'
```
Puppet result:
```
Notice: /Stage[main]/Dns::Server::Config/Concat[/etc/named/named.conf.local]/File[/etc/named/named.conf.local]/content: 
--- /etc/named/named.conf.local 2016-09-22 13:44:53.789000000 +0300
+++ /tmp/puppet-file20160922-43760-19ei9vl      2016-09-22 13:45:34.260000000 +0300
@@ -12,3 +12,13 @@
 acl slave {
   10.7.99.2/32;
 };
+zone "Example.com" {
+type master;
+    file "/var/named/db.Example.com";
+
+};
+zone "example.com" {
+type master;
+    file "/var/named/db.example.com";
+
+};

Notice: /Stage[main]/Dns::Server::Config/Concat[/etc/named/named.conf.local]/File[/etc/named/named.conf.local]/content: content changed '{md5}cda654d9a74f3b01c9a43f2fb341ba66' to '{md5}a00c11bdc620ecd62d8755271e05e82c'
Info: Concat[/etc/named/named.conf.local]: Scheduling refresh of Class[Dns::Server::Service]
Info: Class[Dns::Server::Config]: Scheduling refresh of Class[Dns::Server::Service]
Info: Class[Dns::Server::Service]: Scheduling refresh of Service[named]
Error: Could not start Service[named]: Execution of '/bin/systemctl start named' returned 1: Job for named.service failed because the control process exited with error code. See "systemctl status named.service" and "journalctl -xe" for details.
Error: /Stage[main]/Dns::Server::Service/Service[named]/ensure: change from stopped to running failed: Could not start Service[named]: Execution of '/bin/systemctl start named' returned 1: Job for named.service failed because the control process exited with error code. See "systemctl status named.service" and "journalctl -xe" for details.
Notice: /Stage[main]/Dns::Server::Service/Service[named]: Triggered 'refresh' from 1 events
Info: Class[Dns::Server::Service]: Unscheduling all events on Class[Dns::Server::Service]
Info: Stage[main]: Unscheduling all events on Stage[main]
```
This situation is not possible if define dns::zone have only lowercase titles.

For twin reverse zones we have puppet error: "Duplicate declaration: Exec[bump-168.192.in-addr.arpa-serial] is already declared in file..."
https://github.com/ajjahn/puppet-dns/pull/179/files#diff-2e38170d1d92d7eafada3bbc9bf83100R385

solarkennedy:
I would accept PRs that downcased more things, but in a separate PR. The code in this PR is too much to solve, what in my opinion, is a "non problem". Puppet and most filesystems are case-sensitive, but zones are not. My answer to this problem is "Don't do it", as in, don't define 2 of the same zones with different case.

PR #180, opened 2016-09-18
Rework `zone.pp` to create new `dns::zone` resource for `reverse` zones

jearls:
In order to simplify and standardize the creation of dns zones, and to make it easier to verify the existence of dns zones in other resources, when the `reverse` parameter is set to a non-`false` value, a new `dns::zone` resource is created with the proper zone name.

In order to not break existing installations, the new resource makes use of a new `zone_file` parameter to pass in the zone filenam based on the original non-reversed zone name.  The `zone_file` parameter is validated to ensure it's in the dns data directory and begins with `db.`.

For example, `dns::zone { '10.23.45': reverse => reverse }` will result in a new `dns::zone` resource looking something like this:

```
dns::zone { '45.23.10.in-addr.arpa':
    reverse => false,
    zone_file => '.../db.10.23.45'
}
```

The `...` in the `zone_file` parameter will be replaced by the actual path to the zone file, parameterized in `$dns::server::params::data_dir`.

In addition, any `dns::zone` resource that contains uppercase letters in the zone name will have a new `dns::zone` resource created in which the zone name is all lowercase.  For example, `dns::zone { 'Example.COM': }` would result in a new `dns::zone` resource looking something like this:

```
dns::zone { 'example.com':
    zone_file => '.../db.Example.COM'
}
```

Finally, added `spec` tests against the `zone_file` parameter validation.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -222,6 +222,8 @@ define dns::zone (
   }
 
   if $reverse {
+    # validate that $zone consists of 1 to 3 valid IP octets followed by in-addr.arpa
+    validate_re($zone, '^((25[0-5]|2[0-4][0-9]|[0-1]?[0-9][0-9]?)\.){1,3}in-addr.arpa$', 'With reverse != false, zone names must consist of 1 to 3 dot (.) separated numbers in the range 0 - 255.')
     dns::zone { $zone:
       ensure          => $ensure,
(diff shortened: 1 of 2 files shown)


solarkennedy:
Sorry, I'm going to close this as "too complicated" for not enough gain in functionality.

Issue #181, opened 2016-10-24
`dns::server::defaults` class fails with puppetlabs-stdlib 4.10.0


jearls:
In `pupetlabs-stdlib` version 4.10.0, the `validate_re` function was "improved" to include not allowing its use on any value other than a string.  This breaks existing code that relied on `undef` being treated the same as an empty string in the previous version.

I will submit a PR to fix the failing `validate_re` instances.

PR #182, opened 2016-10-24
Ensure `validate_re` calls are wrapped in `if` checks to avoid passing undef

jearls:


jearls:
Fixes #181

solarkennedy:
This means we no longer get protection from things like users accidentally putting in: `enable_sdb => false`, because it will be false in the if block and skip the validation. (right?)

jearls:
reworked to specifically check for `undef`.  better?

solarkennedy:
Looks great! Thank you.

PR #185, opened 2016-11-14
feat query_log - optional parameter  query_log_enable to enable query log

eumel8:


solarkennedy:

PR #192, opened 2017-01-20
2nd chance: feat query_log - optional parameter query_log_enable to enable query log

eumel8:


templates/named.conf.options.erb
```diff
@@ -127,3 +127,25 @@ statistics-channels {
     inet <%= @statistic_channel_ip %> port <%= @statistic_channel_port %><% if @statistic_channel_allow -%> allow { <%= @statistic_channel_allow.join('; ') %>; }<% end -%>;
 };
 <% end -%>
+
+<% if @query_log_enable -%>
+logging {
+
+        channel query_logging {
+                file "/var/cache/bind/named_querylog"
```
solarkennedy: Is there a variable you can use (`data_dir`) to remove some of the hard-coded pathiness?

eumel8: ${data_dir} is on Debian systems '/etc/bind/zones', I wouldn't use this for log file targetting. A new variable is required for that.

eumel8:
Don't know why Travis failed to build on Ruby 1.8. Maybe an error in the build system. Is there an option to re-trigger the build?

solarkennedy:
It has a tendency to break because it is no longer really supported.
I'll (probably) end up removing it from the matrix.

PR #193, opened 2017-01-24
fixing path issue which prevents working when path is not /etc/bind

ppouliot:
Currently the config_dir part of this was set to /etc/bind.   Changed to module variable config_dir for proper execution.


solarkennedy:
@ajjahn @jearls ready to drop support for ruby 1.8.7? It is broken again.

ajjahn:
@solarkennedy Yes.

PR #196, opened 2017-01-27
Permit overriding global `forwarders` with an empty list (per zone)

jjthiessen:
This PR:
* Expands on the zone types that are permitted to have `forward` and `forwarders` directives ([ref](https://ftp.isc.org/isc/bind9/cur/9.9/doc/arm/Bv9ARM.ch06.html#zone_statement_grammar))
* Allows for overriding global `forwarders` with an empty list on a per zone basis
* Sneaks in some reformatting/re-indenting of `templates/zone.erb` for consistency and cleaner looking generated zone files


jjthiessen:
This addresses https://github.com/ajjahn/puppet-dns/issues/204 (like https://github.com/ajjahn/puppet-dns/pull/210), but would require a major version bump due to backwards incompatible behaviours.

Issue #197, opened 2017-02-02
Incorrect data_dir on Ubuntu


apeschel:
It appears that the dns::server class has the data_dir hard coded to `/etc/bind/zones`. However, this is not the correct directory for dynamic zone files. The correct path on Ubuntu should be `/var/lib/bind` or `/var/lib/bind/zones`. It would be nice if this was configurable, or at least changed to the correct location.

apeschel:
I figured out how to properly set this value via hiera with no required changes. Just a rookie mistake on my part.

(dns::server::config::data_dir in hiera)

apeschel:
After attempting to actually switch the zone, I found that it just wasn't practical to do. None of the dns::record sub-classes allow you to override the data_dir parameter.

apeschel:
To clarify on *why* this is a problem, apparmor does not allow writing in `/etc/bind`. If you are using a zone with DDNS, then you need to use `/var/lib/bind` for your zones.

apeschel:
Fixed in https://github.com/ajjahn/puppet-dns/pull/200

PR #200, opened 2017-02-10
BREAKING (for debian): Fix default Ubuntu data_dir

apeschel:
This PR allows the data_dir field to be set for all record defined types. It also allows parameters to be passed in to the dns::server class for convenience.

It also fixes the problem with the default Ubuntu data_dir by switching it to /var/lib/bind/zones. The tests have been updated to reflect this change.

Finally, some issues around the testing were found while trying to get it to work locally. Some fixes have been provided to help with this.


apeschel:
Related to https://github.com/ajjahn/puppet-dns/issues/197

manifests/record/a.pp
```diff
@@ -8,21 +8,32 @@
   $data,
   $ttl = '',
   $ptr = false,
-  $host = $name ) {
+  $host = $name,
+  $data_dir = $::dns::server::params::data_dir,
```
solarkennedy: All of these `params` references need to be `dns::server::config::foo`

params just is a class to provide defaults. Users are encouraged to set thing, but not via params, via the config class.

apeschel: @solarkennedy I can make the change, but every other parameterized class and reference to a parameter in this module is done directly with `dns::server::params`. It would be nice to have them switched to `dns::server::config` as well.

apeschel:
@solarkennedy I've updated the PR to use `dns::server::config` would appreciate a re-review when you have time.

solarkennedy:
Can you address the style issues to make travis green?
https://travis-ci.org/ajjahn/puppet-dns/jobs/215670156

apeschel:
@solarkennedy Updated

PR #201, opened 2017-02-10
Remove ${host}.${zone} split in record::ptr::by_ip

reynir:
NOTE: Breaking change.
record::ptr::by_ip is a convenient function, but the ${host}.${zone}
split is not always a convenient or meaningful distinction to make when
creating PTR records. This commit removes this distinction from
record::ptr::by_ip and rewrites record::a to do the string
concatenation.

This is a twin pull request. The other is a backwards compatible version.

### DIFF
--- a/manifests/record/a.pp
+++ b/manifests/record/a.pp
@@ -20,9 +20,9 @@ define dns::record::a (
   }
 
   if $ptr == 'all' {
-    dns::record::ptr::by_ip { $data: host => $host, zone => $zone }
+    dns::record::ptr::by_ip { $data: host => "${host}.${zone}" }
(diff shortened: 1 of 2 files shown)


solarkennedy:
cc @jearls

jearls:
I have no objections to this.  This type was created as a way to do iteration over multiple IPs in an A record without relying on Puppet 4's iteration syntax.  However, as this _is_ a breaking change (just in case someone used the type on their own), would this require a major version bump?

reynir:
I would say it would require a major version bump - while it's not documented in the README it's very easy to find and is documented in the file - I was personally (trying to) use it.

jearls:
This PR should be addressed by PR #207, now merged, correct?

PR #202, opened 2017-02-10
Add record::ptr::by_ip_new

reynir:
record::ptr::by_ip is a convenient function, but the host.zone split is
not always a convenient or meaningful distinction to make when creating
PTR records. record::ptr::by_ip_new does not make this distinction.
This commit rewrites record::ptr::by_ip to use record::ptr::by_ip_new in
a trivial way.

This is a twin pull request. The other is a compatibility breaking version.

### DIFF
--- a/manifests/record/ptr/by_ip.pp
+++ b/manifests/record/ptr/by_ip.pp
@@ -74,38 +74,10 @@ define dns::record::ptr::by_ip (
   $ttl = undef ,
   $ip = $name ) {
 
-  # split the IP address up into three host/zone pairs based on class A, B, or C splits:
-  # For IP address A.B.C.D,
-  # class C => host D / zone C.B.A.IN-ADDR.ARPA
-  # class B => host D.C / zone B.A.IN-ADDR.ARPA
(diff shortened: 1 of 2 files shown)


reynir:
Sorry, this PR seems to be rubbish. I will stare at it some more when I'm less tired.

solarkennedy:
cc @jearls

jearls:
FYI - The error in the travis log means that there's a puppet lint problem - probably from the mis-aligned `=>` in the invocation of `dns::record::ptr::by_ip_new`.

However, I have an idea for a simpler way to implement this in a way that's still compatible - basically, make the `zone` parameter to `dns::record::ptr::by_ip` optional, and only append it (with the preceding `.`) if it's defined.

Would this satisfy your use case?

- Johnson

reynir:
Yea, that sounds like a fine solution.

jearls:
@reynir if #207 provides equivalent and satisfactory functionality, would you yield this PR and #201 in favor of #207?

reynir:
Yea, sure.

PR #207, opened 2017-03-02
`dns::record::ptr::by_ip`: Permit `zone` parameter to be empty

jearls:
... to allow fully-qualified `host` values.

Also create basic spec tests for dns`dns::record::ptr::by_ip` type

### DIFF
--- a/manifests/record/ptr/by_ip.pp
+++ b/manifests/record/ptr/by_ip.pp
@@ -10,13 +10,17 @@
 # The ip address for the ptr record.  Defaults to the resource title.
 #
 # * `$host`
-# The (unqualified) host pointed to by this ptr record.
+# The host pointed to by this ptr record.  If `$zone` is a non-empty
+# domain, it will be appended to the value of `$host` in the `PTR`
+# record; if `$zone` is undefined or empty, then `$host`
+# must include the domain name (but must *not* include any trailing `.`).
 #
 # * `$ttl`
 # The TTL of the records to be created.  Defaults to undefined.
 #
 # * `$zone`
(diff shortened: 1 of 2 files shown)


jearls:
This is an alternate proposal for #202

manifests/record/ptr/by_ip.pp
```diff
@@ -102,10 +104,16 @@
...
+  } else {
+    $fqdn = $host
+  }
+
   dns::record::ptr { "${octet}.${reverse_zone}":
     host => $octet,
     zone => $reverse_zone,
     ttl  => $ttl ,
-    data => "${host}.${zone}"
+    data => $fqdn ,
```
solarkennedy: Can you remove all these ` ,` (extra spaces before the commas)
I'm not sure there is a specific style guide rule to that, but it just looks kinda different to me

manifests/record/ptr/by_ip.pp
```diff
@@ -10,13 +10,16 @@
 # The ip address for the ptr record.  Defaults to the resource title.
 #
 # * `$host`
-# The (unqualified) host pointed to by this ptr record.
+# The host pointed to by this ptr record.  This must be fully qualified
```
solarkennedy: Can you clarify if there is supposed to be a period or not?

jearls:
Okay, I fixed the comments for both `$host` and `$zone` to indicate *no* trailing `.`.

I also cleaned up the spaces-before-commas.  I use that in my own modules because, to me, it's much more readable.  Most of the time, I remember not to do that when I'm working on other people's code.  Sorry about that.

Issue #203, opened 2017-02-15
working_dir not expanding in template


lancherider:
In the named.conf.options.erb template, the working_dir variable is not expanding. So, this line:

`file "<%= @working_dir %>/named_querylog"`

ends up looking like this in /etc/named/named.conf.options:

`file "/named_querylog"`

Interestingly, using `@data_dir` performs the substitution just fine, and of course, explicitly setting the path works too.

PR #208, opened 2017-03-04
Add `working_dir` parameter to `dns::server::options`

jearls:
... to fix the named.conf.options expansion.  Also
validate `data_dir` and `working_dir` as absolute paths,
and add `data_dir` and `working_dir` tests to the spec
test for `dns::server::options`.


jearls:
Addresses #203

Issue #204, opened 2017-02-24
Cannot set empty forwarders for zone


apeschel:
It's not possible to override the forwarders for a zone to be empty. This means that authoritative zones will use the default forwarders from the options file even if this is not the desired behavior.

jearls:
I can think of three ways to fix this:
1. Add a new boolean parameter, something like `force_empty_forwarders` or something, which will tell the zone to include the `forwarders { }` list even if it's empty
2. Change the default value of the `allow_forwarders` to `undef`, which would mean no `forwarders { }` list is defined in the zone, and change the behavior of an empty `allow_forwarders` array to create an empty `forwarders { }` list in the zone.  This is a breaking change and would require a major version bump.
3. Add a new `forwarders` parameter that has the semantics described above, and deprecate the existing `allow_forwarders` parameter.  If both are specified, they would be merged together; if only `allow_forwarders` is specified, it would have the same semantics as the current parameter.  Deprecating the parameter requires a minor version bump; removing the deprecated parameter later would require a major version bump.

@solarkennedy @ajjahn any preferences?  I sort of lean towards deprecating the current parameter and introducing a new parameter with the desired semantics.

solarkennedy:
@jearls I agree on #3.

PR #210, opened 2017-03-06
Add `forwarders` to `dns::zone`; deprecate `allow_forwarder`

ghost:
This change adds the `forwarders` parameter to `dns::zone`, deprecating
`allow_forwarder` in favor of `forwarders`.  `forwarders` extends
`allow_forwarder` by allowing both an *empty* forwarders list as well as
an *undefined* forwarders list (the former meaning that the zone does not
forward, while the latter leaves the global forwarders option in place):

> [*forwarders*]
> An array of IP addresses and optional port numbers to which queries
> for this zone will be forwarded (based on the *forward_policy*
> setting).  If the optional port number is included, it must be
> separated from the IP address by the word `port` - for example,
> `[ '192.168.100.102 port 1234' ]`.  If passed an empty array or the
> boolean value `false`, the zone will not forward.  If passed `true`
> or left undefined, the zone will use the global forwarders defined
> in `dns::server::options`.
> *Note* - this parameter deprecates and should be used in place of
> the *allow_forwarder* parameter.  If both parameters are passed in,
> only *forwarders* will take effect.

This addresses Issue #204

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -47,12 +47,14 @@
 #   the array that matches the requester's address will be used.
 #
 # [*allow_forwarder*]
+#   > **DEPRECATED** in favor of the `forwarders` parameter.  This
+#   > parameter will be removed in the next major release.
 #   An array of IP addresses and optional port numbers to which queries
 #   for this zone will be forwarded (based on the *forward_policy*
 #   setting).  If the optional port number is included, it must be
 #   separated from the IP address by the word `port` - for example, `[
 #   '192.168.100.102 port 1234' ]`.  Defaults to an empty array, which
-#   means no forwarding will be done.
+#   means the global forwarders options will be used.
 #
 # [*allow_query*]
 #   An array of IP addresses from which queries should be allowed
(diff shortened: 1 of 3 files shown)


jearls:
Note - as this PR deprecates a parameter, SemVer requires a minor version bump if/when this is published.

manifests/zone.pp
```diff
@@ -175,12 +189,46 @@
...
+      if $forwarders {
+        $zone_forwarders = undef
+      } else {
+        $zone_forwarders = []
+      }
+    } else {
+      validate_array($forwarders)
+      $zone_forwarders = $forwarders
+    }
+  } else {
```
apeschel: Not sure about the style on this module, but this would read cleaner for me if lines 224-225 were replaced with `elsif`.

apeschel (changes requested):
Looks good, just one style suggestion.

jjthiessen:
I have a similar change in https://github.com/jjthiessen/puppet-dns/commit/1e2d3357b089b0610c654301cd4b943ae56be928, except I went with [option 2](https://github.com/ajjahn/puppet-dns/issues/204#issuecomment-284132399) and extended forwarders to _slave_ and _stub_ zones (see the [zone statement grammar](https://ftp.isc.org/isc/bind9/cur/9.9/doc/arm/Bv9ARM.ch06.html#zone_statement_grammar) for justification).

If this (_option 3_) is the preferred approach, can we augment _templates/zone.erb_ to support _slave_ and _stub_ zones and get this merged?

If it's preferable to keep the empty forwarder feature addition, and _slave_ and _stub_ forwarder support in separate PRs, I can submit a follow-up PR (it likely makes sense to merge this first as (either way) one will have to be rebased on the other).

Is there any particular hold-up on this PR?

jjthiessen:
As an aside, https://github.com/jjthiessen/puppet-dns/commits/add-forwarder-support-for-slave-and-stub-zones is this branch with the practically trivial addition of slave and stub forwarder support.

PR #215, opened 2017-04-12
Fix breaking tests on old puppet and ruby

jearls:
Restrict puppetlabs-concat version to < 2.2.1

The puppetlabs-concat module version `2.2.1` (just released) introduces a breaking change for Puppet < `3.2.0`: it includes the use of `Puppet::Parameter::Boolean` for type checking, but those parameter types were not introduced until `3.2.0`.

----

Restrict puppetlabs_spec_helper gem version to < 2.1.1

puppetlabs_spec_helper gem version `2.1.1` introduces a dependency on the `parallel` gem, which requires Ruby 2.0.  This breaks all testing on older Ruby versions.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,6 +1,8 @@
 fixtures:
   forge_modules:
     stdlib: "puppetlabs/stdlib"
-    concat: "puppetlabs/concat"
+    concat:
+      repo: "puppetlabs/concat"
+      ref: "2.2.0"
   symlinks:
     dns: "#{source_dir}"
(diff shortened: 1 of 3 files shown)


jearls:
FYI, [MODULES-4687](https://tickets.puppetlabs.com/browse/MODULES-4687) submitted against `puppetlabs-concat` breaking change, and [MODULES-4688](https://tickets.puppetlabs.com/browse/MODULES-4688) submitted against `puppetlabs_spec_helper` breaking change.

PR #217, opened 2017-04-14
Add support for response-policy

HielkeJ:


mcblum:
@ajjahn it would be amazing to merge this! I was looking for how to do this tonight.

mcblum:
@HielkeJ hey - I'm somewhat new to Puppet. Do you happen to know if I can use the version that you opened a PR for?

HielkeJ:
@mcblum: I have used this almost a year ago. I can't think of a reason it should not work anymore.

mcblum:
@HielkeJ yeah man, I tried it last night and it's working great! Ironically, apparently, RPZ doesn't actually solve our problem of needing a zone for `example.com` but needing to allow `images.example.com` to be resolved externally, but that's a different issue :)

PR #219, opened 2017-05-25
Add extra_options parameter to bind::server::options

amateo:
So you can use more options than the explicitily defined

### DIFF
--- a/manifests/server/options.pp
+++ b/manifests/server/options.pp
@@ -144,6 +144,10 @@
 #   The working directory where the query log will be stored.
 #   Default: `/var/cache/bind` in Debian, `${data_dir}/data` in RedHat
 #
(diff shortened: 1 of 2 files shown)


manifests/server/options.pp
```diff
@@ -219,6 +224,8 @@
   validate_absolute_path($data_dir)
   validate_absolute_path($working_dir)
 
+  validate_hash($extra_options)
```
solarkennedy: this is causing the tests to fail, maybe default to `{}`?

amateo: You are write. I have fixed it.

solarkennedy:
Can you rebase against master now?

amateo:
Rebased

PR #220, opened 2017-05-25
Add parameters to configure bind logging

amateo:
The query_log_enable makes a predefined configuration. This commit adds
log_channels and log_categories so you could make a more fine
configuration.

### DIFF
--- a/templates/named.conf.options.erb
+++ b/templates/named.conf.options.erb
@@ -159,4 +159,22 @@ logging {
         category default { syslog_errors;  };
         category lame-servers { null; };
 };
+<%- elsif !@log_channels.empty? or !@log_categories.empty? -%>
+logging {
+  <%- @log_channels.each do |channel, options| -%>
+    channel <%= channel %> {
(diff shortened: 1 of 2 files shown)


templates/named.conf.options.erb
```diff
@@ -153,4 +153,24 @@ logging {
         category default { syslog_errors;  };
         category lame-servers { null; };
 };
+<%- else -%>
```
solarkennedy: Are you sure this should be an `else` ? 

If a user doesn't enable `query_log_enable`, then the log_channel stuff should activate?

I expected this to be additional things *inside* the `logging {` block

amateo: It is a `else` with an `if` inside. Maybe I should write it with an `elsif`.
The idea is that if you enable `query_log_enable`, then you get the default configuration, but if you don't *and* you configure `log_categories` and `log_channels` you could set a more fine logging configuration

amateo:
I have changed the code, using now an `elsif`.

PR #221, opened 2017-05-25
Add controls parameter to bind::server::options

amateo:
This parameter is a hash of options lines to configure controls in bind


amateo:
I'm going to make a new PR with control configured in the same way than statistics.

PR #223, opened 2017-05-26
Fix controls channel statement

amateo:
The statement to declare controls channel is "controls" not
"controls-channel".


amateo:
I'm sorry but the commit of PR #222 had an errata in the template. I have fixed it.

Issue #35, opened 2013-10-24
Add support for views


danzilio:
I really want to implement views, but after hacking at it for awhile, I'm not entirely sure how to implement this. The problem is that zones need to be nested inside the view in which they belong. There's a one-to-many relationship between views and zones, and the puppet language doesn't give us a good way implement this relationship. I've just started toying with the idea of implementing this by creating a namespace for the concat priority. An example would be the view having a concat priority of 1000, but using an inline template to make the concat priority of the zones inside that view be 1001. I'm really not sure what this would look like...I'm going to have to flesh it out some more. Any ideas?

PR #224, opened 2017-05-26
Add dns::server::view

amateo:
Add `dns::server::view` to configure dns views.
In order to handle this, I have also modified a little the zone implementation to be able to use a file different than _named.conf.local_ to include the zone configurations.
It also adds a `enable_default_zones` to `dns::server` so the inclussion of the `named.conf.default-zones` file could be ommited when views are used (in this case all zones must be defined inside a view, so that file could not be globally included).

### DIFF
--- a/files/named.conf.default-zones
+++ b/files/named.conf.default-zones
@@ -1,0 +1,30 @@
+// File managed with puppet
+// prime the server with knowledge of the root servers
+zone "." {
+	type hint;
+	file "/etc/bind/db.root";
+};
+
+// be authoritative for the localhost forward and reverse zones, and for
+// broadcast zones as per RFC 1912
+
+zone "localhost" {
+	type master;
+	file "/etc/bind/db.local";
+};
+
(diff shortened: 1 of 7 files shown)


solarkennedy:
This fixes #35.

solarkennedy:
Can you check out the failing tests? You are using advanced parsing language and this module is designed to support older versions of puppet.

solarkennedy:
@amateo I've reviewed all this and I don't have any objections or comments. 
You will probably be the first user of this feature, so my only recommendation is that if you want to make sure it is not broken in the future, add some tests on the view define, as there is quite a bit of logic in the erb that could go wrong.

But I'm ready to ship. This is a large change with the default zone stuff, so I'll probably bump the version to indicate that. Let me know if you are good to go.

amateo:
Perfect.
I'll make some tests in another PR, if you don't mind.

solarkennedy:
sgtm

PR #225, opened 2017-07-10
Support Debian 9

neomilium:


jearls:
this allows the module to be used, but does not provide any access to new features provided by debian9's bind version 9.10, nor does it address any possible incompatibilities with the older bind 9.9.  should those changes be addressed in separate pull requests?

neomilium:
@jearls you're right. The goal of this PR is only to fix module usage after upgrade and have been reported to work as it. It does not add support for additional features. In short term, I have no plan to improve this module more than this fix.

solarkennedy:
I'm ok with "allowing" this module to work on debian 9 so at least the incompatibilities can be discovered.

PR #227, opened 2017-07-16
Find a version of rspec-puppet that works with older puppets

jearls:
The `rspec-puppet` gem update to version `2.6.0` broke things for Puppet 3.0 - 3.5.  Until this gets fixed (I've got a PR into the `rspec-puppet` github repo for the fix), this change will limit `rspec-puppet` to earlier than `2.6.0`.


solarkennedy:
Thank you!

PR #231, opened 2017-08-28
Fix concat::fragment usage.

amateo:
Last versions of concat module have removed the ensure parameter for
concat::fragment (deprecated since v2.0.0).


solarkennedy:
Can you adjust the tests that are looking out for this `ensure` param?

amateo:
Fixed.

Issue #230, opened 2017-07-26
Split Horizon


mattslater:
Hi,

I'm having issues with implementing split horizon with this module. Can you confirm that it's possible and if so would you mind showing me a little example?

I'm at a stage where im having problems with the hash for the zone within view.pp 
$zones                = {}, 

A point in the right direction would be great.

Thanks!

jearls:
I'm not familiar with the `dns::server::view` implementation, as that was contributed by someone else - but it appears that the `$zones` parameter is a hash of the arguments you would pass to `dns::zone` - for example, if you wanted to have the `example.com` domain in two different views, `internal` and `external`, and prevent zone transfers in the `external` view, you might do something like this:

    dns::server::view { 'internal':
        # (view-specific parameters)...
        zones => {
            'example.com' => {
                # (other zone parameters to define the zone)...
                allow_transfer => 'any',
            },
        }
    }
    dns::server::view { 'external':
        # (view-specific parameters)...
        zones => {
            'example.com' => {
                # (other zone parameters to define the zone)...
                allow_transfer => 'none',
            },
        }
    }

mattslater:
Hi Jearls,

Thanks for you help, that worked, although It doesn't look like this module will work for split horizon.

amateo:
Hi Mattslater,

I'm the coder of the `dns::server::view implementation`. It is used as Jearls pointed and I'm afraid it doesn't implement split horizon, but I don't know if you could use it to _manually_ configured (manually creating the appropriates views and zones).

mattslater:
Hi @amateo 

Thanks for taking the time to reply, I have just picked this up again. I will probably end up just configuring them manually :)

PR #233, opened 2017-09-09
Add `zone` parameter to Dns::Zone

ghost:
This is a step towards a split-horizon DNS server (#230).
by allowing the same DNS zone to be defined multiple
times using different resource titles.

Currently, this will cause problems with the `ptr`
parameter of `dns::record::a`, so it's not yet a
complete solution.


templates/zone.erb
```diff
@@ -1,4 +1,6 @@
-zone "<%= @zone %>" {
+; BIND definition for <%= @_zone %> zone
```
rmlandvreugd: This comment type appears wrong.
Shouldn't it be `//` ?

PR #234, opened 2017-09-25
Fix "left match operand must be string" error

tprobinson:
I'm unsure if it's a Puppet version thing, but this line was giving me the following error:

Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Left match operand must result in a String value. Got an Integer. at /modules/dns/manifests/record.pp:19:6 at /modules/dns/manifests/record/a.pp:17 on node puppetmaster

Forcing the match operand to a string fixed it for me.

### DIFF
--- a/manifests/record.pp
+++ b/manifests/record.pp
@@ -16,7 +16,9 @@ define dns::record (
 
   $zone_file_stage = "${data_dir}/db.${zone}.stage"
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
You must fix the linting errors
https://travis-ci.org/ajjahn/puppet-dns/jobs/279600783

You will also need `# lint:ignore:disable_only_variable_string`

PR #240, opened 2017-10-26
Check zones before using them

tprobinson:
These changes validate all zone files with named-checkzone before putting the new zone file in or restarting BIND. I was having a lot of trouble with making mistakes in my records leading to my DNS servers crashing, this should prevent that.

Tested in CentOS 7 and Ubuntu Xenial as working, though I'm not familiar enough with Rspec to write a test for it unfortunately.

### DIFF
--- a/manifests/server/params.pp
+++ b/manifests/server/params.pp
@@ -19,9 +19,9 @@ class dns::server::params {
       $default_dnssec_enable     = true
       $default_dnssec_validation = 'auto'
       if versioncmp( $::operatingsystemmajrelease, '8' ) >= 0 {
-        $necessary_packages = ['bind9']
+        $necessary_packages = [ 'bind9', 'bind9utils' ]
       } else {
-        $necessary_packages = [ 'bind9', 'dnssec-tools' ]
+        $necessary_packages = [ 'bind9', 'bind9utils', 'dnssec-tools' ]
       }
     }
     'RedHat': {
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -252,7 +252,7 @@ define dns::zone (
       mode    => '0644',
       replace => $zone_replace,
       require => Class['dns::server'],
-      notify  => Exec["bump-${zone}-serial"]
+      notify  => Exec["test-${zone}"]
     }
     concat::fragment{"db.${name}.soa":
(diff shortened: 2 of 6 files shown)


jearls:
i think the tests for this would not be unit rspec tests, but would be in system acceptance tests - which i don't know how to write yet.

I think this is a good change.  Can you give an example of the relevant portions of the puppet run output when the `named-checkzone` test fails?

jearls:
I just thought of something:  Does it make a difference one way or the other to make the `named-checkzone` test part of the `bump serial` exec itself, using the `onlyif` parameter, instead of having a separate exec?

        exec { "bump-${zone}-serial":
          command     => "sed '8s/_SERIAL_/${zone_serial}/' ${zone_file_stage} > ${zone_file}",
          path        => ['/bin', '/sbin', '/usr/bin', '/usr/sbin'],
          refreshonly => true,
          provider    => posix,
          user        => $dns::server::params::owner,
          group       => $dns::server::params::group,
          onlyif      => "named-checkzone ${zone} ${zone_file_stage}",
          require     => Class['dns::server::install'],
          notify      => Class['dns::server::service'],
        }

i think both would be similar, but i don't know if the behaviors would be identical, and i don't know if one way would be preferred over the other.

tprobinson:
The problem with doing it in `onlyif`  is that it doesn't represent a failure if the command returns a non-zero code, it's more like "My work is already done". So if a bad zone file was generated and `named-checkzone` failed, it wouldn't replace the zone file with `bump serial` or restart the service, but it wouldn't notify the puppet run that anything had gone wrong either. It's quite easy to create a bad zone file unfortunately, so we definitely want to tell the user if that happens.

I don't have logs on hand, but something that kept happening (because I was migrating zone files from an older DNS server into Puppet) was making CNAME and A records with the same host. The zone file would be generated and the DNS server would be restarted, then crash with the unhelpful "CNAME and other data" error, causing lots of problems. With these changes, the "CNAME and other data" error appears in the exec fail instead.

jearls:
Yeah, that makes sense.

tprobinson:
I've been using this fork of the module in our production environment succesfully, these changes make it more reliable. Can this PR please be reviewed and merged?

PR #246, opened 2018-11-08
moving line into the define case due to failure of puppet when line is outside

ppouliot:
Currently fails on newer versions of puppet.   Moving this line inside of the defines allows for proper processing.


barnumbirr:
@ajjahn @jearls @solarkennedy Any update on getting this PR merged?

